PR #4, opened 2018-08-02
global: REANA DB extracted from REANA Commons

diegodelemos:


tiborsimko:
Perhaps add 

```
Co-authored-by: XXX YYY <xxx.yyy@cern.ch>
```

in the commit log message, including each individual author from `r-commons`, in other to better highlight the code move?

tiborsimko:
(BTW, speaking of the git commit log message, perhaps we should use `(connect reanahub/reana#71)`  in the commit log itself, not only on the github issue, so that a trace would be left in the source code repository too; just like for `references` and `closes` and other such directives) should work just as well...)

PR #7, opened 2018-08-07
models: increase `prettified_cmd` size

### LINKED ISSUE #6: models: prettified_cmd too small for big commands/scripts such as atlas-recast/statanalysis
Stemmed from https://github.com/reanahub/reana-demo-atlas-recast/pull/10#issuecomment-410991766.

diegodelemos:
* Examples such as `reana-demo-atlas-recast` use big commands and
  when they are extended (i.e. all palceholders in a yadage step
  are replaced which the actual values) they can be big as a bash
  script (closes #6).

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -28,7 +28,7 @@ import enum
 import uuid
 
 from sqlalchemy import (Boolean, Column, DateTime, Enum, ForeignKey, Integer,
-                        String, UniqueConstraint)
+                        String, Text, UniqueConstraint)
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import relationship
 from sqlalchemy_utils import JSONType, UUIDType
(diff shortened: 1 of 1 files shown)


reana_db/models.py
```diff
@@ -182,7 +182,7 @@ class Job(Base, Timestamp):
     max_restart_count = Column(Integer)
     deleted = Column(Boolean)
     logs = Column(String, nullable=True)
-    prettified_cmd = Column(String(1000))
+    prettified_cmd = Column(Text)
```
dinosk: Should we change `cmd` to `Text` as well?

diegodelemos: I think you are right, at first I thought that there wouldn't be any problem because variables are not expanded, but taking into account [how `scripts` can be defined in `yadage`](https://yadage.readthedocs.io/en/latest/definingworkflows.html?highlight=script#process-definitions) we better change it as well.

PR #10, opened 2018-09-17
models: reana_specification field in Workflow

dinosk:
* Adds reana_specification field in Workflow table.
    
* Adds operational_parameters field to contain parameters related to
  the workflow engine execution.
    
Connects https://github.com/reanahub/reana-db/issues/12


reana_db/models.py
```diff
@@ -122,6 +129,14 @@ def get_workspace(self):
         """
         return build_workspace_path(self.owner_id, self.id_)
 
+    def get_parameters(self):
+        """Return workflow parameters."""
+        return self.reana_specification['workflow'].get('parameters', {})
```
diegodelemos: Shouldn't this be:
```suggestion
        return self.reana_specification.get('parameters', {})
```

PR #13, opened 2018-10-18
release: 0.4.0.dev20181018

diegodelemos:


diegodelemos:
Yes, as long as is it is a pre-release we won't fill it, once we are sure that everything is stable for 0.4.0 we tag it and release.

PR #14, opened 2018-10-18
models: fix get_parameters() for empty inputs

dinosk:
* Fixes Workflow.get_parameters() for the case of workflows without
  inputs.

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -131,7 +131,7 @@ class Workflow(Base, Timestamp):
 
     def get_parameters(self):
         """Return workflow parameters."""
-        return self.reana_specification['inputs'].get('parameters', {})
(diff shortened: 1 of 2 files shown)


reana_db/models.py
```diff
@@ -131,7 +131,10 @@ def get_workspace(self):
 
     def get_parameters(self):
         """Return workflow parameters."""
-        return self.reana_specification['inputs'].get('parameters', {})
+        parameters = {}
+        if 'inputs' in self.reana_specification:
+            parameters = self.reana_specification['inputs'].get('parameters')
```
diegodelemos: Why not just
```suggestion
            parameters = self.reana_specification.get('inputs', {}).get('parameters', {})
```

PR #15, opened 2018-11-02
models: addition of deleted WorkflowStatus

dinosk:
Connects https://github.com/reanahub/reana-client/issues/192

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -60,6 +60,7 @@ class WorkflowStatus(enum.Enum):
     running = 1
     finished = 2
     failed = 3
+    deleted = 4
 
 
 class Workflow(Base, Timestamp):


reana_db/models.py
```diff
@@ -60,6 +60,7 @@ class WorkflowStatus(enum.Enum):
     running = 1
     finished = 2
     failed = 3
+    deleted = 4
```
diegodelemos: Can you please add `stopped` already? (according to https://github.com/reanahub/reana-client/issues/192#issuecomment-437003151 we should be fine with those status for now). And also rebase.

dinosk: I think it's better to add it in the PR group for `workflow stop`, if it is part of this PR, it won't show for example that the `workflow stop` adds something to the DB.

diegodelemos: Agree 👍 can you rebase?

PR #26, opened 2019-02-08
workflow: add get owner access token

diegodelemos:
Connects reanahub/reana-workflow-controller#184.


reana_db/models.py
```diff
@@ -144,6 +144,14 @@ def get_specification(self):
         """Return workflow specification."""
         return self.reana_specification['workflow'].get('specification', {})
 
+    def get_owner_access_token(self):
+        """Return workflow specification."""
```
roksys: """Return workflow owner access token."""

PR #56, opened 2020-01-22
models: adds restart and restart_info columns

roksys:


reana_db/models.py
```diff
@@ -100,6 +100,8 @@ class Workflow(Base, Timestamp):
     run_stopped_at = Column(DateTime)
     run_number = Column(Integer)
     job_progress = Column(JSONType, default=dict)
+    restart = Column(Boolean, default=False)
+    restart_info = Column(JSONType, default=list)
```
tiborsimko: As discussed briefly, while developing workflows, people may stay on the same run number for a long time.  This leads to multiple workflow restarts from various steps, which means that the restart column would grow... It may be better to rather extend on orthogonal direction, i.e. not new columns but rather new rows/tables.  Might be more scalable.

Also, if we have a "runtime log" table that stores all the runs, then making statistics how much CPU time user Y consumed would be easy for both runs and reruns, since the script can operate on both the same way.  (If the rerun info is stored in a column, the script would have to operate on both rows and columns, making it more error-prone).

We could think of a (i) "workflow runs" table that would store all the runs and reruns; this will also help to make the main workflow table short in case it grows too much.  (Kind of like keeping inbox vs archive folders.)

We could think of a more general (ii) "audit" table that would store not only run and rerun information, but other various actions of interest for taking note of.  The table in this case would be rather general (=who performed what action at what time with some free parameters e.g. runtime cost and disk space cost). 

We can muse more IRL on those options.

PR #57, opened 2020-01-27
models: adds workspace_path column

roksys:
* Saves workspace path column in order to allow reuse of
  workspace.

* Addresses reanahub/reana-workflow-controller/issues/289

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -13,9 +13,10 @@ from __future__ import absolute_import
 import enum
 import uuid
 
-from sqlalchemy import (Boolean, Column, DateTime, Enum, ForeignKey, Integer,
+from sqlalchemy import (Boolean, Column, DateTime, Enum, Float, ForeignKey,
                         String, Text, UniqueConstraint)
 from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import relationship
 from sqlalchemy_utils import JSONType, UUIDType
 from sqlalchemy_utils.models import Timestamp
@@ -98,8 +99,10 @@ class Workflow(Base, Timestamp):
     run_started_at = Column(DateTime)
     run_finished_at = Column(DateTime)
     run_stopped_at = Column(DateTime)
-    run_number = Column(Integer)
+    _run_number = Column('run_number', Float)
     job_progress = Column(JSONType, default=dict)
+    workspace_path = Column(String)
+    restart = Column(Boolean, default=False)
     # job_progress = {
     #  jobs_total = {total: job_number}
     #  jobs_running = {job_ids: [], total: c}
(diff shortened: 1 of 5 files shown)


reana_db/models.py
```diff
@@ -154,6 +158,8 @@ def __init__(self,
             self.run_number = 1
         else:
             self.run_number = last_workflow.run_number + 1
+        self.workspace_path = workspace_path or self.get_workspace()
```
diegodelemos: Only if no one else is calling `get_workspace` anymore, we could just call `build_workspace_path` directly and get rid of `get_workspace` method:
```suggestion
        self.workspace_path = workspace_path or build_workspace_path(owner_id, id_)

```

I think it would be more clear, read as "if there is no workspace path, one will be built", WDYT?

diegodelemos: Can't we apply the previous comment diff, just as it is? That would allow us to completely get rid of `get_workspace`, I don't see it being used anywhere:
```console
$ rg "get_workspace\(\)"
reana-db/reana_db/models.py
161:        self.workspace_path = workspace_path or self.get_workspace()
```

reana_db/models.py
```diff
@@ -150,22 +155,21 @@ def __init__(self,
             name=name,
             owner_id=owner_id).\
             order_by(Workflow.run_number.desc()).first()
-        if not last_workflow:
-            self.run_number = 1
+        if restart:
```
diegodelemos: An observation regarding how the run number is exposed to users. If you run a workflow once, you will see:
```console
$ rc run 
[INFO] Creating a workflow...
workflow.1.0
[INFO] Uploading files...
File code/helloworld.py was successfully uploaded.
File data/names.txt was successfully uploaded.
[INFO] Starting workflow...
workflow.1.0 is running
```
It seems a bit weird to see the `1.0` when I haven't used the restart feature. I would rather expect to see ` 1` and then when I run `reana-client start --restart -w workflow.1` I would expect to see `1.1`.

I think we could easily do this here by using:
```python
In [1]: run_number_with_restart = 1.1 

In [2]: run_number_with_no_restart = 2.0                                    

In [3]: run_numbers = [run_number_with_restart, run_number_with_no_restart] 

In [4]: for run_number in run_numbers:
   ...:     if run_number.is_integer(): 
   ...:         print(int(run_number))
   ...:     else: 
   ...:         print(run_number) 
   ...:            
1.1
2
```

If we do it, I would try to move the whole `run_number_increase` logic to a different method, so we keep the constructor tidy.

roksys: Thanks for suggestions, it was addressed.

tests/test_models.py
```diff
@@ -0,0 +1,69 @@
...
+"""REANA-DB models tests."""
+
+from uuid import uuid4
+
+import pytest
+
+from reana_db.models import Workflow
+
+
+def test_workflow_run_number_assignment(db, session):
```
diegodelemos: 🚀

PR #60, opened 2020-02-05
Workflow query fix

roksys:


reana_db/utils.py
```diff
@@ -103,7 +103,7 @@ def _get_workflow_with_uuid_or_name(uuid_or_name, user_uuid):
         # `run_number` was specified.
         # Check `run_number` is valid.
         try:
-            float(run_number)
+            run_number = float(run_number)
```
tiborsimko: Hmm, looking at instances of float and int casting in the code base, and looking at possible RDBMS differences in handling floats, what about:

- either forcing the run number to be `float` everywhere on the cluster side (e.g. `42.0`) and hiding the `42.0` complexity into `42` on the  reana-client side?

- or using two integer columns, run_number and rerun_number, in the DB model, each properly indexed for speed, and defining proper getter/setter hybrid property for run_number lookups that would look in both columns based on run_number string without any needs for float casting?

It might be more robust that way, i.e. considering run number as a string composed of two integers separated by dot, and not casting anything. Let's muse IRL?

PR #64, opened 2020-04-22
models: add audit log and user token tables

### LINKED ISSUE #18: Addition of audit table.
Actions like deletion of a workflow and others, could be logged in a more complete way if there would be a table to store meta information about them.
This could be an `audit` table with the following structure:
```
  audit
----------
creation_timestamp
last_modification_timestamp
user (id)
action (text)
details (json)

```

In the workflow deletion case a row would show:
who (user) 
did (deleted in action)
what (workflow id in json)
and how (hard/soft delete details in json).

Other actions like file uploads, downloads, removals could be captured in a similar way.

mvidalgarcia:
closes #18
closes reanahub/reana-ui#93

#### TODO:
- [x] Docstrings
- [x] `type` column and adapt
- [x] DB constraint: one single `active` token per `type`.
- [x] Tests

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -43,6 +43,7 @@ class User(Base, Timestamp):
     full_name = Column(String(length=255))
     username = Column(String(length=255))
     workflows = relationship("Workflow", backref="user_")
+    audit_logs = relationship("AuditLog", backref="user_")
 
     def __repr__(self):
         """User string represetantion."""
@@ -301,3 +302,25 @@ class JobCache(Base, Timestamp):
     result_path = Column(String(1024))
     workspace_hash = Column(String(1024))
     access_times = Column(JSONType)
+
+
+class AuditLogAction(enum.Enum):
+    """Enumeration of audit log actions."""
+
+    request_token = 0
+
+
+class AuditLog(Base, Timestamp):
+    """Audit log table."""
+
+    __tablename__ = 'audit_log'
+
+    id_ = Column(UUIDType, unique=True, primary_key=True,
+                 default=generate_uuid)
+    user_id = Column(UUIDType, ForeignKey('user_.id_'), nullable=False)
+    action = Column(Enum(AuditLogAction), nullable=False)
+    details = Column(JSONType)
+
+    def __repr__(self):
+        """Audit log string representation."""
+        return f'<AuditLog {self.id_} {self.action}>'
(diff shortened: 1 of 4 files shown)


reana_db/models.py
```diff
@@ -301,3 +302,26 @@ class JobCache(Base, Timestamp):
...
+class AuditLog(Base, Timestamp):
+    """Audit log table."""
+
+    __tablename__ = 'audit_log'
+
+    id_ = Column(UUIDType, unique=True, primary_key=True,
+                 default=generate_uuid)
+    user_id = Column(UUIDType, ForeignKey('user_.id_'), nullable=False)
+    # created_at = Column(DateTime) # comes from Timestamp (created)
+    # modify_at = Column(DateTime) # comes from Timestamp (updated)
```
mvidalgarcia, line 322: Just for you to know that I didn't create these columns because apparently they come with `Timestamp`.

reana_db/models.py
```diff
@@ -301,3 +302,24 @@ class JobCache(Base, Timestamp):
...
+
+class AuditLog(Base, Timestamp):
+    """Audit log table."""
+
+    __tablename__ = 'audit_log'
+
+    id_ = Column(UUIDType, unique=True, primary_key=True,
+                 default=generate_uuid)
+    user_id = Column(UUIDType, ForeignKey('user_.id_'), nullable=False)
+    action = Column(Enum(AuditLogAction), nullable=False)
```
tiborsimko, line 321: In order to be future-friendly, please foresee adding a `details` JSON column, even if we won't use it much yet.  First possible use case: ask for reason to use REANA via a textarea near "Request token" button with leading text going like "Please explain briefly what would you like to use REANA for" that would be showing for non-physics users.  So the action `details` column would store `{"reason": "foo bar"}`.  See also https://github.com/reanahub/reana-db/issues/18#issuecomment-617660607

tiborsimko:
Since after this PR we'll have a different `User` table with token column gone, please amend also `r-server`'s facility for importing/exporting users, see ` _import_users()` and `_export_users()` there. The MVP for importing/exporting would be to import/export only "active" tokens; we'll use this for upgrading REANA production instance.  (The import/export of all tokens including "revoked" will be a plus for later.)

reana_db/models.py
```diff
@@ -18,6 +18,7 @@
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import relationship
+from sqlalchemy.schema import Index
```
diegodelemos, line 21: Unused?

PR #65, opened 2020-04-28
models: add log action method for users

mvidalgarcia:
closes reanahub/reana-server#235

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -10,5 +10,5 @@ pydocstyle reana_db && \
 isort -rc -c -df **/*.py && \
 check-manifest --ignore ".travis-*" && \
 sphinx-build -qnNW docs docs/_build/html && \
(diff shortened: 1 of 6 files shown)


run-tests.sh
```diff
@@ -10,5 +10,5 @@ pydocstyle reana_db && \
 isort -rc -c -df **/*.py && \
 check-manifest --ignore ".travis-*" && \
 sphinx-build -qnNW docs docs/_build/html && \
-python setup.py test && \
+REANA_SQLALCHEMY_DATABASE_URI=sqlite:///test.db python setup.py test && \
```
diegodelemos, line 13:
```suggestion
REANA_SQLALCHEMY_DATABASE_URI=sqlite:// python setup.py test && \
```
So it doesn't create the file.

PR #73, opened 2020-06-18
cli: add reana-db command

diegodelemos:
* Wraps alembic command to enable database upgrades (closes reanahub/reana#336).

### DIFF
--- a/reana_db/alembic.ini
+++ b/reana_db/alembic.ini
@@ -9,6 +9,8 @@ script_location = %(here)s/alembic
 # directories, initial revisions must be specified with --version-path
 version_locations = %(here)s/alembic/versions
 
+file_template = %%(year)d%%(month).2d%%(day).2d_%%(hour).2d%%(minute).2d_%%(rev)s_%%(slug)s
+
 [post_write_hooks]
 # post_write_hooks defines scripts or Python functions that are run
 # on newly generated revision scripts.  See the documentation for further
--- a/reana_db/alembic.ini
+++ b/reana_db/alembic.ini
@@ -1,0 +1,56 @@
+# A generic, single database configuration.
+
+[alembic]
+# path to migration scripts
+script_location = %(here)s/alembic
+
+# version location specification; this defaults
+# to reana_db/alembic/versions.  When using multiple version
+# directories, initial revisions must be specified with --version-path
+version_locations = %(here)s/alembic/versions
+
+[post_write_hooks]
+# post_write_hooks defines scripts or Python functions that are run
+# on newly generated revision scripts.  See the documentation for further
+# detail and examples
+
+# format using "black" - use the console_scripts runner, against the "black" entrypoint
+hooks=black
+black.type=console_scripts
+black.entrypoint=black
+
+
+# Logging configuration
+[loggers]
+keys = root,sqlalchemy,alembic
+
+[handlers]
+keys = console
+
+[formatters]
+keys = generic
+
+[logger_root]
+level = WARN
+handlers = console
+qualname =
+
+[logger_sqlalchemy]
+level = WARN
+handlers =
+qualname = sqlalchemy.engine
+
+[logger_alembic]
+level = INFO
+handlers =
+qualname = alembic
+
+[handler_console]
+class = StreamHandler
+args = (sys.stderr,)
+level = NOTSET
+formatter = generic
+
+[formatter_generic]
+format = %(levelname)-5.5s [%(name)s] %(message)s
+datefmt = %H:%M:%S
--- a/reana_db/cli.py
+++ b/reana_db/cli.py
@@ -1,0 +1,192 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2020 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""REANA DB command line."""
+
+import os
+
+from alembic import command
+from alembic import config as alembic_config
+import click
+
+from reana_db.database import init_db
+
+
+@click.group()
+def cli():
+    """REANA database commands."""
+
+
+@cli.command()
+def init():
+    """Show REANA database migration recipes history."""
+    init_db()
+    click.secho("Database initialised.", fg="green")
+
+
+@cli.group("alembic")
+@click.pass_context
+def alembic_group(ctx):
+    """REANA database migration commands.
+
+    Note that this command is just a light wrapper around alembic.
+    """
+    reana_alembic_ini = os.path.join(os.path.dirname(__file__), "alembic.ini")
+    config = alembic_config.Config(reana_alembic_ini)
+    ctx.obj = config
+
+
+@alembic_group.command()
+@click.option(
+    "-m", "--message", default=None, help="Message string to use with 'revision'"
+)
+@click.option(
+    "--autogenerate",
+    is_flag=True,
+    help=(
+        "Populate revision script with candidate "
+        "migration operations, based on comparison of "
+        "database to model"
+    ),
+)
+@click.option(
+    "--sql",
+    default=False,
+    help=(
+        "Don't emit SQL to database - dump to standard output/file instead. See alembic docs on offline mode."
+    ),
+)
+@click.option(
+    "--head",
+    default="head",
+    help=("Specify head revision or <branchname>@head to base new revision on."),
+)
+@click.option(
+    "--splice",
+    is_flag=True,
+    help=("Allow a non-head revision as the 'head' to splice onto."),
+)
+@click.option(
+    "--branch-label",
+    default=None,
+    help=("Specify a branch label to apply to the new revision"),
+)
+@click.option(
+    "--version-path",
+    default=None,
+    help=("Specify specific path from config for version file"),
+)
+@click.option(
+    "--rev-id",
+    default=None,
+    help=("Specify a hardcoded revision id instead of generating one"),
+)
+@click.option(
+    "--depends-on",
+    default=None,
+    help=(
+        "Specify one or more revision identifiers which this revision should depend on."
+    ),
+)
+@click.pass_obj
+def revision(
+    config,
+    message,
+    autogenerate,
+    sql,
+    head,
+    splice,
+    branch_label,
+    version_path,
+    rev_id,
+    depends_on,
+):
+    """Create a REANA database alembic revision."""
+    command.revision(
+        config,
+        message=message,
+        autogenerate=autogenerate,
+        sql=sql,
+        head=head,
+        splice=splice,
+        branch_label=branch_label,
+        version_path=version_path,
+        rev_id=rev_id,
+        depends_on=depends_on,
+    )
+
+
+@alembic_group.command()
+@click.argument("revision", default="head")
+@click.option(
+    "--sql",
+    is_flag=True,
+    help=(
+        "Don't emit SQL to database - dump to standard output/file instead. See alembic docs on offline mode."
+    ),
+)
+@click.option(
+    "--tag",
+    default=None,
+    help=("Arbitrary 'tag' name - can be used by custom env.py scripts."),
+)
+@click.pass_obj
+def upgrade(config, revision, sql, tag):
+    """Upgrade REANA database."""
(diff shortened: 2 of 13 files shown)


diegodelemos:
A new command would be added:
```
$ reana-db init
$ reana-db alembic history
$ reana-db alembic upgrade head
$ ...
```

It will be available in any component which installs it. If we don't want this to happen, we can easily refactor it out to either REANA-Server or to a new different module, let me know what you think.

setup.py
```diff
@@ -64,6 +65,7 @@
     author_email="info@reana.io",
     url="https://github.com/reanahub/reana-db",
     packages=["reana_db",],
+    entry_points={"console_scripts": ["reana-db=reana_db.cli:cli"],},
```
tiborsimko, line 68: We'll most probably need also: 

```
    include_package_data=True,
```

to include `alembic ini` and friends.

reana_db/alembic.ini
```diff
@@ -0,0 +1,56 @@
...
+# to reana_db/alembic/versions.  When using multiple version
+# directories, initial revisions must be specified with --version-path
+version_locations = %(here)s/alembic/versions
+
+[post_write_hooks]
+# post_write_hooks defines scripts or Python functions that are run
+# on newly generated revision scripts.  See the documentation for further
+# detail and examples
+
+# format using "black" - use the console_scripts runner, against the "black" entrypoint
```
mvidalgarcia, line 17: When I create a revision I get a `FAILED`, not sure if it comes from black, have you seen it?

```console
root@reana-server-66d5649989-mqcbw:/code# reana-db alembic revision -m 'add lots of things'
  Generating /code/modules/reana-db/reana_db/alembic/versions/270a3a1b9e85_add_lots_of_things.py ...  done
  Running post write hook "black" ...
  FAILED
```

reana_db/cli.py
```diff
@@ -0,0 +1,192 @@
...
+    config = alembic_config.Config(reana_alembic_ini)
+    ctx.obj = config
+
+
+@alembic_group.command()
+@click.option(
+    "-m", "--message", default=None, help="Message string to use with 'revision'"
+)
+@click.option(
+    "--autogenerate",
```
mvidalgarcia, line 49: I was checking this option as I think it can be pretty useful to generate prefilled revision but I get
```console
alembic.util.exc.CommandError: Can't proceed with --autogenerate option; environment script /code/modules/reana-db/reana_db/alembic/env.py does not provide a MetaData object or sequence of objects to the context.
```

I think we need to add MetaData to all the models as stated [here](https://alembic.sqlalchemy.org/en/latest/api/autogenerate.html#getting-diffs)

reana_db/alembic.ini
diegodelemos: No, I didn't see this error and I tested thoroughly the revision creation. The config is the default one, so if you have format on-save I don't see why this would be happening, have you modified the revision template?

reana_db/alembic.ini
```diff
@@ -0,0 +1,56 @@
+# A generic, single database configuration.
```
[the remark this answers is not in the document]
diegodelemos: Sounds good to me, let's do it.

reana_db/cli.py
diegodelemos: If that enables autogenerating migrations let's do it, [here](https://alembic.sqlalchemy.org/en/latest/autogenerate.html) the specific implementation details.

reana_db/alembic.ini
mvidalgarcia: It was failing because I didn't have Black installed in the server pod. This dependency comes from `pytest-reana` which is [`tests_require`](https://setuptools.readthedocs.io/en/latest/setuptools.html) so I guess it's not installed on deploy

reana_db/cli.py
mvidalgarcia: I wonder if this option should be enabled by default, WDYT?

The generated result should look something like this, so one needs to adjust it in any case:

```python
def upgrade():
    # ### commands auto generated by Alembic - please adjust! ###
    op.add_column("user_", sa.Column("test", sa.String(length=255), nullable=True))
    # ### end Alembic commands ###


def downgrade():
    # ### commands auto generated by Alembic - please adjust! ###
    op.drop_column("user_", "test")
    # ### end Alembic commands ###
```

reana_db/database.py
```diff
@@ -23,10 +24,19 @@
 Base.query = Session.query_property()
 
 
+def _before_create(target, connection, **kwargs):
+    """Create schemas as SQLAlchemy doesn't do it and set 'public' schema as default."""
+    CreateSchema("reana").execute(connection)
+    # Otherwise Invenio tables end up in 'reana' schema.
+    sql = "ALTER ROLE {} SET search_path TO public".format(DB_USERNAME)
```
mvidalgarcia, line 31: I couldn't find a better/programmatic solution to run this statement. I'm afraid this can be vulnerable to SQL injection attacks.. although I've tried many `DB_USERNAME` and I didn't succeed e.g `"reana'; SELECT hacked"` and so on.
Do you have any ideas to improve this?

diegodelemos: I know this would be making assumptions, but since we are the ones in charge of the code that instantiates the DBs, what if we invert the order and we move the [Invenio DB creation](https://github.com/reanahub/reana-server/blob/486c96d4968243fff1d3faaba21dc219f2ee20af/scripts/setup#L18) before the REANA DB creation? Wouldn't that solve it?

mvidalgarcia: It works but then an empty revision contains this:

```python
ef upgrade():
    # ### commands auto generated by Alembic - please adjust! ###
    op.create_foreign_key(None, 'audit_log', 'user_', ['user_id'], ['id_'], source_schema='reana', referent_schema='reana')
    op.create_foreign_key(None, 'job_cache', 'job', ['job_id'], ['id_'], source_schema='reana', referent_schema='reana')
    op.create_foreign_key(None, 'user_token', 'user_', ['user_id'], ['id_'], source_schema='reana', referent_schema='reana')
    op.create_foreign_key(None, 'workflow', 'user_', ['owner_id'], ['id_'], source_schema='reana', referent_schema='reana')
    # ### end Alembic commands ###


def downgrade():
    # ### commands auto generated by Alembic - please adjust! ###
    op.drop_constraint(None, 'workflow', schema='reana', type_='foreignkey')
    op.drop_constraint(None, 'user_token', schema='reana', type_='foreignkey')
    op.drop_constraint(None, 'job_cache', schema='reana', type_='foreignkey')
    op.drop_constraint(None, 'audit_log', schema='reana', type_='foreignkey')
    # ### end Alembic commands ###
```

mvidalgarcia: Check [`quote_ident`](https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.quote_ident) to escape `DB_USERNAME`.

run-tests.sh
```diff
@@ -6,9 +6,39 @@
...
+OLD="$(docker ps --all --quiet --filter=name=reana-postgres)"
+if [ -n "$OLD" ]; then
+  echo 'Cleaning old DB container...' && \
+  docker stop reana-postgres
+fi
+echo 'Starting DB container...' && \
+docker run --rm --name reana-postgres -p 5432:5432 -e POSTGRES_PASSWORD=mysecretpassword -d postgres && \
+check_ready "Postgres" _db_check && \
+python setup.py test && \
+echo 'Stopping DB container...' && \
```
tiborsimko, line 42: Since the script has now grown and we now have separate functions, can you please amend it as follows:

- create functions for everything
- introduce newlines between them
- use echo statements everywhere (e.g. 'Checking Black compliance...') for consistency
- the container name is indeed called "postgres", but the DB name is "PostgreSQL", better use that one
- we can use '==> [INFO]' and '==> [ERROR]' for messages that we create ourselves

tests/test_models.py
```diff
@@ -11,29 +11,25 @@
 from uuid import uuid4
 
 import pytest
-import sqlalchemy
-from mock import patch
```
tiborsimko, line 15: Can you look also at `conftest.py` that probably does not need mock patch either?

reana_db/database.py
mvidalgarcia: Changing schema name solves this problem, we don't need `ALTER ROLE` anymore.

run-tests.sh
```diff
@@ -6,9 +6,39 @@
...
 # under the terms of the MIT License; see LICENSE file for more details.
 
+export REANA_SQLALCHEMY_DATABASE_URI=postgresql+psycopg2://postgres:mysecretpassword@localhost/postgres
+
+# Verify that db container is running before continuing
+check_ready() {
+    RETRIES=40
+    while ! $2
+    do
+        echo "Waiting for $1, $((RETRIES--)) remaining attempts..."
```
tiborsimko, line 16: We either need to change shell from `sh` to `bash`, or else amend POSIX compliance:

```
$ shellcheck ./run-tests.sh 

In ./run-tests.sh line 16:
        echo "Waiting for $1, $((RETRIES--)) remaining attempts..."
                                        ^-- SC2039: In POSIX sh, -- is undefined.


In ./run-tests.sh line 26:
_db_check(){ docker exec --user postgres reana-postgres bash -c "pg_isready" &>/dev/null; }
                                                                             ^---------^ SC2039: In POSIX sh, &> is undefined.

For more information:
  https://www.shellcheck.net/wiki/SC2039 -- In POSIX sh, &> is undefined.

```console

File: run-tests.sh [hunk 9]
```diff
@@ -1,14 +1,63 @@
...
+}
+
+_db_check() {
+    docker exec --user postgres reana-postgres bash -c "pg_isready" &>/dev/null;
+}
+
+clean_old_db_container() {
+    OLD="$(docker ps --all --quiet --filter=name=reana-postgres)"
+    if [ -n "$OLD" ]; then
+        echo '==> [INFO] Cleaning old DB container...' && \
```
tiborsimko, line 33: We don't need `&&` in these internal statements after `echo`.  It is sufficient if `&&` were enchained only at the bottom of the script when calling them.

Actually, it would be simpler to delete `&&` everywhere, since we can do "stop-after-first-failure" by setting:
```
# Quit on errors
set -o errexit
```

in the beginning of the script.

P.S. We could also add the following for shell script security:

```
# Quit on unbound symbols
set -o nounset
```

at the top as well.

run-tests.sh
```diff
@@ -1,14 +1,69 @@
...
+
+_db_check() {
+    docker exec --user postgres postgres__reana-db bash -c "pg_isready" &>/dev/null;
+}
+
+clean_old_db_container() {
+    OLD="$(docker ps --all --quiet --filter=name=postgres__reana-db)"
+    if [ -n "$OLD" ]; then
+        echo '==> [INFO] Cleaning old DB container...'
+        docker stop postgres__reana-db
```
tiborsimko, line 40: BTW is it enough to `stop` the old container? Shouldn't we also `rm` it, just to be sure that it is created anew and does not carry over any past state?  The current thing is obviously OK for Travis CI runs, where everything is run on a blank slate VM, but if one uses `run-tests.sh` script locally, then one ends with a lot of stopped (and never removed) local containers that could carry over some unwanted state. Perhaps we should do `rm` instead of `stop` here and also at the end of tests, to end up with a blank slate again?

(This applies also to the other SQLite->PostgreSQL PRs.)

mvidalgarcia: `docker run` has already `--rm` so it's removed automatically when stopped.

reana_db/version.py
```diff
@@ -14,4 +14,4 @@
 
 from __future__ import absolute_import, print_function
 
-__version__ = "0.7.0.dev20200619"
+__version__ = "0.7.0.dev20200708"
```
tiborsimko, line 17: Can you please remove the commit `release: 0.7.0.dev20200708` commit and rather add a new one on the branch head that would be `0.7.0a1` version already?

Question: (a) Are we ready to start tagging these alpha releases properly? Which would mean to update release notes, to mint github tag, etc? (b) Or are we OK to still not tag these properly?  Which means just to release alpha1 on PyPY but not separating release notes and not minting git tags? I think the latter option (b) might do for speed purposes, since alphas are still meant to be pretty much internal.

diegodelemos: I would go for (b) as it is more gradual.

PR #78, opened 2020-08-05
installation: adds autodiscovered packages

### LINKED ISSUE #77: installation: missing subpackages in production build
Our `setup.py` configuration [doesn't dynamically discover new subpackages](https://github.com/reanahub/reana-db/blob/ce07a07ec6f5bb667fb20ee66aca48fa68f61836/setup.py#L67). We should [use `find_packages()`](https://github.com/reanahub/reana-db/blob/ce07a07ec6f5bb667fb20ee66aca48fa68f61836/setup.py#L50) so subpackages are included in the production build.
```console
$ pip show reana-db
Name: reana-db
Version: 0.7.0a2
Summary: REANA-DB.
Home-page: https://github.com/reanahub/reana-db
Author: REANA
Author-email: info@reana.io
License: UNKNOWN
Location: /Users/rodrigdi/.virtualenvs/test/lib/python3.6/site-packages
Requires: psycopg2-binary, alembic, reana-commons, SQLAlchemy, sqlalchemy-utils, cryptography
Required-by: 
$ ls /Users/rodrigdi/.virtualenvs/test/lib/python3.6/site-packages/reana_db
__init__.py cli.py      database.py utils.py
__pycache__ config.py   models.py   version.py
```
_`alembic`, a subpackage present in master, is not shipped with the current production builds._

diegodelemos:
* Adds missing subpackge `alembic` (and all future subpackages) to
  the production builds (closes #77).


setup.py
```diff
@@ -60,11 +60,12 @@
     name="reana-db",
     version=version,
     description=__doc__,
+    license="MIT",
```
diegodelemos, line 63: @tiborsimko, addressing https://github.com/reanahub/reana-db/issues/77#issuecomment-669292448: When the package license is only present in classifiers `pip show` doesn't show it. For now the only workaround is to use the `license` parameter, however, the [official docs](https://packaging.python.org/specifications/core-metadata/?highlight=license#license) say that this field should only be used to declare custom licenses. More detailed information https://github.com/pypa/pip/issues/6677

tiborsimko: Hmm, so there is no good solution right now... Anyway, if adding `license=MIT` does not break anything apparent, only goes a bit against the spirit of official docs, we could perhaps have it everywhere while releasing later alpha versions...

PR #82, opened 2020-09-04
database: create schema only if it doesn't exist

### LINKED ISSUE #81: schema: create schema only if it doesn't exist
At the moment of deploying REANA `0.7.0-alpha.2` we have found out that:
1. The REANA user has no permissions to create schemas by default
2. We will create the schema with the admin user, however, [current implementation is not prepared to expect the schema to be already created](https://github.com/reanahub/reana-db/blob/506f1d8fcc61cdc84febbb8dbd3c11156a0861ed/reana_db/database.py#L31).

Actions:
- [x] Make the code resilient, so if the scheme already exists the creation step will be skipped.

To move to another issue: 
- [ ] Document the procedure for installing REANA with external Postgresql databases
   - [ ] By granting the db user with create schema priviledges
   - [ ] By creating the schema manually with an admin user

diegodelemos:
* Closes #81 .

### DIFF
--- a/reana_db/database.py
+++ b/reana_db/database.py
@@ -28,7 +28,9 @@ def init_db():
     """Initialize the DB."""
     import reana_db.models
 
-    event.listen(Base.metadata, "before_create", CreateSchema("__reana"))
+    reana_schema_name = "__reana"
+    if not engine.dialect.has_schema(engine, reana_schema_name):
+        event.listen(Base.metadata, "before_create", CreateSchema(reana_schema_name))
     if not database_exists(engine.url):
         create_database(engine.url)
     Base.metadata.create_all(bind=engine)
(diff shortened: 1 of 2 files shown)


reana_db/database.py
```diff
@@ -28,7 +28,9 @@ def init_db():
     """Initialize the DB."""
     import reana_db.models
 
-    event.listen(Base.metadata, "before_create", CreateSchema("__reana"))
+    reana_schema_name = "__reana"
+    if not engine.dialect.has_schema(engine, reana_schema_name):
+        event.listen(Base.metadata, "before_create", CreateSchema("__reana"))
```
mvidalgarcia, line 33:
```suggestion
        event.listen(Base.metadata, "before_create", CreateSchema(reana_schema_name))
```

tiborsimko (approved):

PR #84, opened 2020-09-22
models: add disk usage methods for User and Workflow models

audrium:
closes https://github.com/reanahub/reana-server/issues/279

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -34,6 +34,8 @@ from sqlalchemy_utils.types.encrypted.encrypted_type import AesEngine
 from reana_db.config import DB_SECRET_KEY
 from reana_db.utils import build_workspace_path
 
+from reana_commons.utils import get_disk_usage
+
 Base = declarative_base()
 
 
@@ -120,6 +122,15 @@ class User(Base, Timestamp):
         :return: Path to the user's workspace directory.
         """
         return build_workspace_path(self.id_)
+
+    def get_user_disk_usage(self, block_size=None):
+        """Retrieve user disk usage information."""
+        try:
+            workspace_path = self.get_user_workspace()
+            disk_usage = get_disk_usage(workspace_path, True, block_size)
+            return disk_usage[0]["size"]
+        except Exception:
+            return -1
 
     def request_access_token(self):
         """Create user token and mark it as requested."""
@@ -386,6 +397,10 @@ class Workflow(Base, Timestamp):
     def get_full_workflow_name(self):
         """Return full workflow name including run number."""
         return "{}.{}".format(self.name, str(self.run_number))
+
+    def get_workspace_disk_usage(self, summarize=False, block_size=None):
+        """Retrieve disk usage information of a workspace."""
+        return get_disk_usage(self.workspace_path, summarize, block_size)
 
     @staticmethod
     def update_workflow_status(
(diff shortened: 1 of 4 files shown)


reana_db/models.py
```diff
@@ -387,6 +398,18 @@ def get_full_workflow_name(self):
...
+    def get_workspace_disk_usage(self, summarize=False, block_size=None):
+        """Retrieve disk usage information of a workspace."""
+        disk_usage_info = get_disk_usage(self.workspace_path, summarize, block_size)
+        # create pairs of (size, filename)
+        filesize_pairs = list(zip(disk_usage_info[::2], disk_usage_info[1::2]))
+        filesizes = []
+        for filesize_pair in filesize_pairs:
+            size, name = filesize_pair
+            # trim workspace path in every file name
+            filesizes.append({"name": name[len(self.name) :], "size": size})
```
diegodelemos, line 410: Since this logic is specific to `du`'s output, maybe we can do it inside `get_disk_usage`? Wha do you think of something like this:
- Always returning a list of dicts with `name` and `value`
  - In case the of `summarize==True` it can be a single item list
  - In case of `summarize==False` it can be the list we build here

diegodelemos: (in that case, we would need to improve [`get_disk_usage`](https://github.com/reanahub/reana-commons/pull/208/files#diff-c652280f4f14efe4c398627ef0fb5da5R193) docstring so it is clearly documented)

audrium: I've separated this peace of functionality from the `get_disk_usage` since it's not needed for the `get_user_disk_usage` (User model). But you are right, having this code inside the Workflow model does not make too much sense. I will amend it with your suggestions

reana_db/models.py
```diff
@@ -121,6 +123,15 @@ def get_user_workspace(self):
         """
         return build_workspace_path(self.id_)
 
+    def get_user_disk_usage(self, summarize=True, block_size=None):
+        """Retrieve user disk usage information."""
+        try:
+            workspace_path = self.get_user_workspace()
+            disk_usage = get_disk_usage(workspace_path, summarize, block_size)
+            return disk_usage[0]["size"]
```
diegodelemos, line 131: There is just a little detail that needs to be amended: if someone passes a different value than the default to `summarize`, `False`, they will still be receiving just the first element instead of everything... we have two options:
1. Remove the `summarize` parameter and hardcode its value to `True` inside the function. Assuming that for users we want just the overall usage
2. Have some logic which will change the return value depending on the value of `summarize`

In my opinion (1.) makes more sense as I see no use case in which we want a detailed `du` for all workflows of a user (potentially hundreds of workflows times the number of files they contain in their workspace...), what do you think?

audrium: Option (1) makes more sense to me as well. I've updated the code to accordingly

audrium:
closes https://github.com/reanahub/reana-server/issues/279

diegodelemos:
Could you:
- Drop a line in the CHANGES.rst (as in https://github.com/reanahub/reana-commons/pull/209/files). **[1]**
- Bump REANA-Commons version to `0.8.0a1`, it should soon be out

**[1]**: Sorry I've forgotten about this when reviewing REANA-Commons PR. Every time we do substantial change to the code base of a package we try to update the `CHANGES.rst` so when we release we already have a digest of all that has changed.

diegodelemos (approved):

PR #86, opened 2020-09-24
models: add Quota models

### LINKED ISSUE #83: models: prepare DB schema to hold quota usage
Enrich models to hold quota usage information for HDD and CPU usage.

1. Research how this is being done in other systems:
    - Quotas in HTCondor see [here](https://research.cs.wisc.edu/htcondor/manual/v7.6/condor_userlog.html)
    - Quotas in Kubernets see [here](https://kubernetes.io/docs/concepts/policy/resource-quotas/)
    - Quotas in Slurm see [here](https://slurm.schedmd.com/sreport.html)

      Note: the sense of quota might vary between backends, for instance, in HTCondor it is what the job actually consumed, however for Kubernetes it is the limit the job can reach.

2. Extend models (probably create a new accounting table)
    - MVP workflows and sessions/notebooks, nice to have for jobs.

      Note: Focus on adding every field we can potentially fill, even if we do not fill them in the first version of user quotas implementation (to minimize DB models changes).

diegodelemos:
Closes #83.

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -20,6 +20,7 @@ from sqlalchemy import (
     Enum,
     Float,
     ForeignKey,
+    BigInteger,
     String,
     Text,
     UniqueConstraint,
@@ -492,3 +493,66 @@ class AuditLog(Base, Timestamp):
     def __repr__(self):
         """Audit log string representation."""
         return "<AuditLog {} {}>".format(self.id_, self.action)
+
+
+class ResourceType(enum.Enum):
+    """Enumeration of resource types."""
+
+    cpu = 0
+    gpu = 1
+    disk = 2
+
+
+class ResourceUnit(enum.Enum):
+    """Enumeration of resource usage units."""
+
+    bytes_ = 0
+    milliseconds = 1
+
+
+class Resource(Base, Timestamp):
+    """Resource table."""
+
+    __tablename__ = "resource"
+    __table_args__ = {"schema": "__reana"}
+
+    id_ = Column(UUIDType, primary_key=True, default=generate_uuid)
+    name = Column(String(1024))
+    type_ = Column(Enum(ResourceType), nullable=False)
+    unit = Column(Enum(ResourceUnit), nullable=False)
+    title = Column(String(1024))
+
+    def __repr__(self):
+        """Resource string representation."""
+        return "<Resource {}>".format(self.id_)
+
+
+class UserResource(Base, Timestamp):
+    """User Resource table."""
+
+    __tablename__ = "user_resource"
+    __table_args__ = {"schema": "__reana"}
+
+    user_id = Column(UUIDType, ForeignKey("__reana.user_.id_"), primary_key=True)
+    resource_id = Column(UUIDType, ForeignKey("__reana.resource.id_"), primary_key=True)
+    quota_limit = Column(BigInteger())
+    quota_used = Column(BigInteger())
+
+    def __repr__(self):
+        """User Resource string representation."""
+        return "<UserResource {} {}>".format(self.user_id, self.resource_id)
+
+
+class WorkflowResource(Base, Timestamp):
+    """Workflow Resource table."""
+
+    __tablename__ = "workflow_resource"
+    __table_args__ = {"schema": "__reana"}
+
+    workflow_id = Column(UUIDType, ForeignKey("__reana.workflow.id_"), primary_key=True)
+    resource_id = Column(UUIDType, ForeignKey("__reana.resource.id_"), primary_key=True)
+    quantity_used = Column(BigInteger())
+
+    def __repr__(self):
+        """Workflow Resource string representation."""
+        return "<WorkflowResource {} {}>".format(self.workflow_id, self.resource_id)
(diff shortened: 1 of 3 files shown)


reana_db/models.py
```diff
@@ -492,3 +493,70 @@ class AuditLog(Base, Timestamp):
...
+
+
+class WorkflowResource(Base, Timestamp):
+    """Workflow Resource table."""
+
+    __tablename__ = "user_resource"
+    __table_args__ = {"schema": "__reana"}
+
+    workflow_id = Column(UUIDType, ForeignKey("__reana.workflow.id_"), nullable=False)
+    resource_id = Column(UUIDType, ForeignKey("__reana.user_.id_"), nullable=False)
```
audrium, line 555: Small typo:
```diff
-resource_id = Column(UUIDType, ForeignKey("__reana.user_.id_"), nullable=False)
+resource_id = Column(UUIDType, ForeignKey("__reana.resource_.id_"), nullable=False)
```

reana_db/models.py
```diff
@@ -492,3 +493,70 @@ class AuditLog(Base, Timestamp):
...
+        """UserResource string representation."""
+        return "<User {} consumed {} of {} out of {}>".format(
+            self.user_id, self.quota_used, self.resource_id, self.quota_limit
+        )
+
+
+class WorkflowResource(Base, Timestamp):
+    """Workflow Resource table."""
+
+    __tablename__ = "user_resource"
```
audrium, line 551:
```diff
-__tablename__ = "user_resource"
+__tablename__ = "workflow_resource"
```

reana_db/models.py
```diff
@@ -492,3 +493,70 @@ class AuditLog(Base, Timestamp):
...
+    __table_args__ = {"schema": "__reana"}
+
+    workflow_id = Column(UUIDType, ForeignKey("__reana.workflow.id_"), nullable=False)
+    resource_id = Column(UUIDType, ForeignKey("__reana.user_.id_"), nullable=False)
+    quantity_used = Column(BigInteger())
+
+    def __repr__(self):
+        """Resource string representation."""
```
audrium, line 559:
```diff
-"""Resource string representation."""
+"""WorkflowResource string representation."""
```

reana_db/models.py
```diff
@@ -492,3 +493,78 @@ class AuditLog(Base, Timestamp):
...
+    )
+    resource_id = Column(
+        UUIDType, ForeignKey("__reana.resource.id_"), primary_key=True, nullable=False
+    )
+    quota_limit = Column(BigInteger())
+    quota_used = Column(BigInteger())  # bigint allows us to go up to to 8192 pebibytes
+
+    def __repr__(self):
+        """User Resource string representation."""
+        return "<User {} consumed {} of {} out of {}>".format(
```
mvidalgarcia, line 547: I wonder if this is a good representation of this model. I'm very used to see `<ModelName ....`, from a first glance it might be confused with a `<User>` object.

reana_db/models.py
```diff
@@ -492,3 +493,78 @@ class AuditLog(Base, Timestamp):
...
+    """Resource table."""
+
+    __tablename__ = "resource"
+    __table_args__ = {"schema": "__reana"}
+
+    id_ = Column(UUIDType, primary_key=True, default=generate_uuid)
+    name = Column(String(1024))
+    type_ = Column(Enum(ResourceType), nullable=False)
+    unit = Column(Enum(ResourceUnit), nullable=False)
+    title = Column(String(1024))
```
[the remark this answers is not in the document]
tiborsimko: I was thinking `name` being a short descriptive string e.g. "cpu" (that we could also use hard-coded in the application instead of IDs), then `title` being a long human-readable version of the name e.g. "CPU hours at the Foo cluster", so a new `description` column would be actually yet another to add, if we want one.  (BTW, using this logic, `name` could be much shorter column for efficient DB indexing than 1024 string.)

Alternatively, if we never hardcode `cpu`, `gpu` and friends, but we be always looking them up in the application logic already from the get go, e.g. via reana-commons or live DB queries, then it could be sufficient to have only `name` and `description` indeed.  (and no `title`)

reana_db/models.py
diegodelemos: Amended to follow `<ModelName ....>`

ParthS007 (approved):
LGTM 🚀

mvidalgarcia (approved):
Lovely!

```console
root@reana-server-779f65c44b-vfft6:/code# reana-db alembic upgrade
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade  -> bfacc7e3ba22, Quota tables.

root@reana-server-779f65c44b-vfft6:/code# reana-db alembic downgrade base
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running downgrade bfacc7e3ba22 -> , Quota tables.
```

reana_db/models.py
```diff
@@ -492,3 +493,74 @@ class AuditLog(Base, Timestamp):
...
+
+
+class UserResource(Base, Timestamp):
+    """User Resource table."""
+
+    __tablename__ = "user_resource"
+    __table_args__ = {"schema": "__reana"}
+
+    user_id = Column(
+        UUIDType, ForeignKey("__reana.user_.id_"), primary_key=True, nullable=False
```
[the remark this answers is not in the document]
diegodelemos: You are right, updated (FYI It is for sure implicit in the `primary_key` declaration, however, even if after removing it, the alembic recipes still contain the `nullable=False` for the PKs).

PR #87, opened 2020-09-29
quotas: calculate quota usage for workflows

diegodelemos:
Closes reanahub/reana-workflow-controller#332, closes https://github.com/reanahub/reana/issues/415.

### DIFF
--- a/reana_db/cli.py
+++ b/reana_db/cli.py
@@ -9,12 +9,14 @@
 """REANA DB command line."""
 
 import os
+import sys
 
+import click
 from alembic import command
 from alembic import config as alembic_config
-import click
-
+from reana_db.config import DEFAULT_QUOTA_RESOURCES
 from reana_db.database import init_db
+from reana_db.models import Resource, ResourceType, ResourceUnit
 
 
 @click.group()
@@ -197,3 +199,44 @@ def history(config, rev_range, verbose, indicate_current):
     command.history(
         config, rev_range=rev_range, verbose=verbose, indicate_current=indicate_current
     )
+
+
+@cli.group("quota")
+@click.pass_context
+def quota_group(ctx):
+    """REANA DB quota related commands."""
+
+
+@quota_group.command()
+def create_default_resources():
+    """Create default quota resources."""
+    from reana_db.database import Session
+
+    resource_type_to_unit = {
+        ResourceType.cpu: ResourceUnit.milliseconds,
+        ResourceType.disk: ResourceUnit.bytes_,
+    }
+
+    existing_resources = [r.name for r in Resource.query.all()]
+    default_resources = []
+    for type_, name in DEFAULT_QUOTA_RESOURCES.items():
+        if name not in existing_resources:
+            default_resources.append(
+                Resource(
+                    name=name,
+                    type_=ResourceType[type_],
+                    unit=resource_type_to_unit[ResourceType[type_]],
+                    title=f"Default {type_} resource.",
+                )
+            )
+    if default_resources:
+        Session.add_all(default_resources)
+        Session.commit()
+        click.secho(
+            f"Added resources: {[r.name for r in default_resources]}", fg="green"
+        )
+    else:
+        click.secho(
+            "No action to be taken: default resources already exist.", fg="yellow"
+        )
+        sys.exit(1)
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -12,18 +12,25 @@ from __future__ import absolute_import
 
 import enum
 import uuid
+from datetime import datetime
 
+from reana_commons.utils import get_disk_usage
+from reana_db.config import DB_SECRET_KEY, DEFAULT_QUOTA_RESOURCES
+from reana_db.utils import build_workspace_path
+from requests.sessions import session
 from sqlalchemy import (
+    BigInteger,
     Boolean,
     Column,
     DateTime,
     Enum,
     Float,
     ForeignKey,
-    BigInteger,
     String,
     Text,
     UniqueConstraint,
+    event,
+    func,
 )
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.ext.hybrid import hybrid_property
@@ -31,11 +38,6 @@ from sqlalchemy.orm import relationship
 from sqlalchemy_utils import EncryptedType, JSONType, UUIDType
 from sqlalchemy_utils.models import Timestamp
 from sqlalchemy_utils.types.encrypted.encrypted_type import AesEngine
-
-from reana_db.config import DB_SECRET_KEY
-from reana_db.utils import build_workspace_path
-
-from reana_commons.utils import get_disk_usage
 
 Base = declarative_base()
 
@@ -58,6 +60,7 @@ class User(Base, Timestamp):
     tokens = relationship("UserToken", backref="user_", lazy="dynamic")
     workflows = relationship("Workflow", backref="user_", lazy="dynamic")
     audit_logs = relationship("AuditLog", backref="user_")
+    resources = relationship("UserResource", backref="user_")
 
     def __init__(self, access_token=None, **kwargs):
         """Initialize user model."""
@@ -65,6 +68,7 @@ class User(Base, Timestamp):
             setattr(self, k, v)
         if access_token:
             self.access_token = access_token
+        self.initialize_user_quota_limits()
 
     @hybrid_property
     def active_token(self):
@@ -133,6 +137,28 @@ class User(Base, Timestamp):
         except Exception:
             return -1
 
+    def get_user_cpu_usage(self):
+        """Aggregate user CPU usage."""
+        try:
+            from .database import Session
+
+            cpu_resources = [
+                str(res[0])
+                for res in Session.query(Resource.id_)
+                .filter_by(type_=ResourceType.cpu)
+                .all()
+            ]
+            return (
+                Session.query(func.sum(UserResource.quota_used))
+                .filter(
+                    UserResource.resource_id.in_(cpu_resources),
+                    UserResource.user_id == self.id_,
+                )
+                .scalar()
+            )
+        except Exception:
+            return -1
+
     def request_access_token(self):
         """Create user token and mark it as requested."""
         from .database import Session
@@ -168,6 +194,19 @@ class User(Base, Timestamp):
         Session.add(audit_log)
         Session.commit()
         return audit_log
+
+    def initialize_user_quota_limits(self):
+        """Initialize user quota limits."""
+        resources = Resource.query.all()
+        for resource in resources:
+            self.resources.append(
+                UserResource(
+                    user_id=self.id_,
+                    resource_id=resource.id_,
+                    quota_limit=0,
+                    quota_used=0,
+                )
+            )
 
     def __repr__(self):
         """User string represetantion."""
@@ -344,8 +383,6 @@ class Workflow(Base, Timestamp):
 
     @run_number.expression
     def run_number(cls):
-        from sqlalchemy import func
-
         return func.abs(cls._run_number)
 
     def assign_run_number(self, run_number):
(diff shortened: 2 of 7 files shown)


reana_db/models.py
```diff
@@ -435,6 +456,41 @@ def can_transition_to(self, next_status):
...
+        WorkflowStatus.finished,
+        WorkflowStatus.failed,
+    ]:
+        workflow.run_finished_at = datetime.now()
+    elif new_status in [WorkflowStatus.stopped]:
+        workflow.run_stopped_at = datetime.now()
+
+    finished_at = workflow.run_finished_at or workflow.run_stopped_at
+    if workflow.run_started_at and finished_at:
+        RESOURCE = "cpu"
```
diegodelemos, line 472: Following the discussion during stand-up. We need to find a logic way of knowing which exact CPU resource a workflow has been consumed.

For this first implementation, we have taken the decision to calculate the CPU time at the workflow level. This includes:
- Container related overhead:
    - Meeting pod requirements:
       - Pulling images
       - Mounting volumes
       - Mounting config maps
       - etc...
    - Start pod
    - Note: it doesn't include containers teardown
- Workflow engine orchestration overhead
- The sum of all execution times for jobs

Although fine-grained accounting would be very beneficial, we cannot exactly account for each detail yet. Instead, we will have to account for all of it together. 

My suggestion would be to have a "general CPU" (I know the name isn't great, just coming up with one to kick-off the discussion) which would map one to one to the summary shown in the [your profile page](https://github.com/reanahub/reana-ui/pull/134), being a summary of total CPU/processing power usage. Once we get more advanced into what was consumed on each part, this UI wouldn't be enough and we would could have a dedicated page with a breakdown in general and per workflow.

Note: with the current implementation only `ResourceType.cpu` will be aggregated. If this "general CPU" would include GPUs too (which I think it makes sense) maybe the name has to be "more general" and the implementation should be amended.

reana_db/models.py
```diff
@@ -435,6 +456,41 @@ def can_transition_to(self, next_status):
...
+        cpu_resource = Resource.query.filter_by(name=RESOURCE).one_or_none()
+        if cpu_resource:
+            from .database import Session
+
+            workflow_resource = WorkflowResource(
+                workflow_id=workflow.id_,
+                resource_id=cpu_resource.id_,
+                quantity_used=cpu_milliseconds,
+            )
+            user_resource_quota = UserResource.query.filter_by(
```
diegodelemos, line 484: We are assuming for now that the [admins have successfully populated the quota related tables](https://github.com/reanahub/reana/pull/416). For now, if this is not the case, we return `null` as CPU usage. What should we do?
1. Continue
    - Log what has happened, that quota was requested but null was returned as there is missing config
2. Error out
    - Maybe not if the admins are not interested in accounting usage?

tests/conftest.py
```diff
@@ -33,3 +43,80 @@ def new_user(session):
...
+            logs="",
+            status=WorkflowStatus.created,
+        )
+        # start workflow
+        workflow.status = WorkflowStatus.running
+        workflow.run_started_at = now
+        session.add(workflow)
+        session.commit()
+        # simulate time elapsed
+        sleep(time_elapsed_seconds)
```
diegodelemos, line 118: This is not great, I am working on fixing it...

reana_db/models.py
```diff
@@ -133,6 +137,25 @@ def get_user_disk_usage(self, block_size=None):
         except Exception:
             return -1
 
+    def get_user_cpu_usage(self):
```
diegodelemos, line 140: `get_user_disk_usage` should follow the same approach as `get_user_cpu_usage`, just reading the value from DB and calculate it only if it hasn't been calculated before. This will be possible once https://github.com/reanahub/reana-workflow-controller/issues/333 is implemented.

reana_db/models.py
```diff
@@ -133,6 +137,25 @@ def get_user_disk_usage(self, block_size=None):
...
+
+            cpu_resources = [
+                str(res[0])
+                for res in Session.query(Resource.id_)
+                .filter_by(type_=ResourceType.cpu)
+                .all()
+            ]
+            return (
+                Session.query(func.sum(UserResource.quota_used))
+                .filter(UserResource.resource_id.in_(cpu_resources))
```
mvidalgarcia, line 153: Aren't we missing something like `UserResource.user_id == self.id_` to filter by user here?

reana_db/models.py
mvidalgarcia: Maybe "default CPU"?
I guess for now all the workflows can consume this resource. I wonder if we should have a flag to mark a resource of given `ResourceType` as the default? So we avoid the hardcoding... But still, we'd have the problem of knowing if it was run by CPU or GPU...

reana_db/models.py
mvidalgarcia: Perhaps accounting should be optional and only calculate quotas if `Resource`s and `UserResource`s are set. Otherwise, we would be forcing admins to set quota limits per user (unless we let them set some defaults).

reana_db/models.py
tiborsimko: It makes totally sense to calculate this "general CPU". I guess I would call it with something englobing, like "Processor" or better "Processing" or even more completely "Processing time".  

Some of the use cases mention are perfectly normal to get accounted to the user. If User A has large Docker images, REANA needs to pull them, so it is an attribute of the user workflow... Unlike User B that only needs small images.  (If an image is already pulled, that's a sort of "luck" which can be neglected in the zero-order approach.) Ditto for orchestration time: if a User A has pipeline with 10000 parallel steps, and User B with only 100, it is also perfectly normal to count more for user A, since orchestration of these jobs takes time... Another attribute of the user workflow.
Ditto for mounting volumes...  So I wouldn't go into too much granularity here.

reana_db/models.py
tiborsimko: Why don't we have some default quota limit value (say 0) meaning "no quota"? 

If an admin wants to set quotas to users, this would be always some positive number, say 1 sec or 1 KiB. Because to ban people, admins could just revoke the token... No need to "nullify" quota for this.

The accounting, if cheap, could be done always, so that admins know what's happening... But the end user would see "no quota limit" if the limit value is the default one (0).

reana_db/models.py
diegodelemos: Actions discussed after stand-up:
- [ ] Load fixtures with base `Resource`s (similar to the [current bash script](https://github.com/reanahub/reana/pull/416) but more user-friendly, technical detail to be decided)
- [ ] Create entries in `UserResource` when users are created (note that users can be created through UI and CLI). Whether we set default quota to `null` or `0`, to be dediced.

Still to be agreed upon:

> I wonder if we should have a flag to mark a resource of given ResourceType as the default? So we avoid the hardcoding...

This makes a lot of sense. I see two ways of achieving this:
1. Static config: Treating this as "before deployment" static configuration that would be populated when we ingest the fixtures by selecting one of the `Resource`s as default (see https://github.com/reanahub/reana/issues/277). This would technically require a restart of all the pods that read/require the configuration.
2. Dynamic config: create a new table `Configuration` which would look a bit like follows (very early thought, take with a grain of salt):

    | Scope | Name |  Value | Default | Description | Optional |
    | ------ | ------ | ------ | ------- | ------------ | -------- |
    | global | resource:processing_time:default | null | d717a177-6e99-43d9-b978-d35d11e53048 (FK Resource.id_) | Processing time resource to account for when there is no specific resource to assign to. | false |

    This would scale towards the idea of having [admin roles](https://github.com/reanahub/reana-server/issues/285) changing this configuration through the UI. I've added `scope` as the concept that the default processing time resource might changeable in the future by groups admins (admin of: a collaboration, a working group etc..)]

tiborsimko: Do we need a new table? All the resources-under-quota are basically defined in a table already, so why not to reuse that one? E.g. title/type/description etc is already there... Whether REANA instances will populate it by fixtures, or "by hand", shouldn't matter.

The admins adding new resources by a web UI is perhaps looking too much ahead; they would need to modify source code to make these resources registered as needed, which is a complex task... What I thought about as admin UI for MVP are easily editable things, i.e. *keeping* the IDs and names (that we use as a sort of ID) and altering only presentation stuff like title and description, or altering default quota values, etc.... But no adding/deleting rows, as that would require to change the code base to start collecting those new resource values.

diegodelemos: I see no way, with the current implementation, to solve the problem of choosing a specific row in the `Resource` table to account for a given workflow run. Currently, this is hardcoded.

I thought that a way to tackle the problem is to store somewhere a pointer to the row that represents "default resource for processing time" for this specific case, and in general, would be a pointer to default resources (be it GPU, disk etc...). In my eyes, this pointer could be in a config file or in a database table.

mvidalgarcia: Idea: What if we have two new columns in the `Resource` table called `resource_unit` (yes, I know, it's an awful name but I can't find anything better atm) that can be either `ResourceUnitType.processing` or `ResourceUnitType.storage` and `default` (boolean), which represents if that specific resource unit is the default one. Then we can have a constraint that forces that there can only be a unique default=True per `ResourceUnitType`. Would that work? Maybe it's a bit crazy... bear with me

reana_db/models.py
```diff
@@ -435,6 +476,42 @@ def can_transition_to(self, next_status):
...
+    elif new_status in [WorkflowStatus.stopped]:
+        workflow.run_stopped_at = datetime.now()
+
+    finished_at = workflow.run_finished_at or workflow.run_stopped_at
+    if workflow.run_started_at and finished_at:
+        cpu_time = finished_at - workflow.run_started_at
+        cpu_milliseconds = int(cpu_time.total_seconds() * 1000)
+        cpu_resource = Resource.query.filter_by(
+            name=DEFAULT_QUOTA_RESOURCES["cpu"]
+        ).one_or_none()
```
mvidalgarcia, line 496: Just asking for confirmation. The disk quota is going to be calculated nightly (creation of `WorkflowResource`s and update of `UserResource`)? What if the quota is exceeded before calculated?

mvidalgarcia (approved):
Sweet, I see I'm using my _unlimited_ 🌟 quota.

```sql
reana=# select * from __reana.workflow_resource;
          created           |          updated           |             workflow_id              |             resource_id              | quantity_used 
----------------------------+----------------------------+--------------------------------------+--------------------------------------+---------------
 2020-10-02 16:09:14.315133 | 2020-10-02 16:09:14.315142 | 5752b7e7-9fd9-4775-9ef2-c6cdce6a61a8 | bce5e1de-e078-4cb4-b6b2-c14db9281932 |         10034
 2020-10-02 16:18:26.576069 | 2020-10-02 16:18:26.576074 | efc1a4c8-1ffb-4d46-b7d2-fc6308e39a7a | bce5e1de-e078-4cb4-b6b2-c14db9281932 |         27612
(2 rows)

reana=# select * from __reana.user_resource;
          created           |          updated           |               user_id                |             resource_id              | quota_limit | quota_used 
----------------------------+----------------------------+--------------------------------------+--------------------------------------+-------------+------------
 2020-10-02 14:13:32.748617 | 2020-10-02 14:13:32.74862  | 00000000-0000-0000-0000-000000000000 | b6696145-af6e-4ffd-a69f-72f33afa8b92 |           0 |          0
 2020-10-02 14:13:32.74861  | 2020-10-02 16:18:26.575553 | 00000000-0000-0000-0000-000000000000 | bce5e1de-e078-4cb4-b6b2-c14db9281932 |           0 |      37646
```

I left a couple of comments that not necessarily need to be dealt with right now. 👍

reana_db/models.py
diegodelemos: I think that the [daemon](https://github.com/reanahub/reana-workflow-controller/issues/333) should not be aware of the quota limitations because:
- It should be intelligent enough to not calculate the disk usage for something that hasn't changed (sort of caching)
- We will stop users from increasing their quota usage (https://github.com/reanahub/reana-server/issues/282 and https://github.com/reanahub/reana-server/issues/283), so the action that they will most probably take will be deletion of files and workflows [to recover quota](https://github.com/reanahub/reana-ui/issues/129), and in that case we really want to recalculate it and show it.

The problem I see here though is that disk quota won't be recalculated on-demand/live. That has two solutions:
1. Recalculate on every action of type [deletion, overwriting ...], probably very expensive
2. Add a new functionality for users to trigger the recalculation of their disk usage

What do you think? cc @tiborsimko @audrium

reana_db/cli.py
```diff
@@ -197,3 +199,34 @@ def history(config, rev_range, verbose, indicate_current):
...
+
+
+@cli.group("quota")
+@click.pass_context
+def quota_group(ctx):
+    """REANA DB quota related commands."""
+
+
+@quota_group.command()
+def create_default_resources():
```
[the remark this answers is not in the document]
diegodelemos: Improvements described in (1.) done, much better now, thanks!. (2.) already linked in the PR description :)

PR #90, opened 2020-10-05
models: add Interactive Session model

### LINKED ISSUE #89: models: separate workflows and interactive sessions
Currently, interactive sessions of a user are stored together in the workflows in the same table:

```
class Workflow(Base, Timestamp):
    ...
    interactive_session = Column(Text)
    interactive_session_name = Column(Text)
    interactive_session_type = Column(Text)
```

We need to separate them into a new table.

This would allow for (1) having more than one session on a workflow workspace; (2) opening interactive sessions without necessarily having any workflow yet.

The implementation can go by having a `InteractiveSession` table similar `Workflow` table, then link it to the user and optionally to a workflow, perhaps with a ternary `user_workflow_session` table.

Let's design a new data model holding such a structure.

Let's amend current logic that opens workflows to follow this new structure.

Let's write alembic upgrade recipe to make sure upgrades are smooth.

(And allowing the opening of more than one session e.g. regular Jupyter notebook, e.g. spark notebook, i.e. the same jupyter but with a different docker container, can be thought of later.)

audrium:
closes #89


reana_db/models.py
```diff
@@ -286,6 +286,58 @@ class JobStatus(enum.Enum):
     queued = 5
 
 
+class UserWorkflowSession(Base, Timestamp):
+    """User Workflow Session table."""
+
+    __tablename__ = "user_workflow_session"
+    __table_args__ = {"schema": "__reana"}
+
+    user_id = Column(UUIDType, ForeignKey("__reana.user_.id_"), nullable=False)
```
diegodelemos, line 295: Do we need to add the user here? I rather see this table as `WorkflowSession` or `WorkflowInteractiveSession` and the relation to `User` to be simply backtracked from `InteractiveSession` through `owner_id` (I see ternary relationship was mentioned in the original issue, but I have second thoughts).

reana_db/models.py
```diff
@@ -286,6 +286,58 @@ class JobStatus(enum.Enum):
...
+class InteractiveSessionType(enum.Enum):
+    """Enumeration of interactive session types."""
+
+    jupyter = 0
+
+
+class InteractiveSession(Base, Timestamp):
+    """Interactive Session table."""
+
+    __tablename__ = "interactive_session"
```
diegodelemos, line 317: Not to be addressed in this PR:

Addressing the following line in https://github.com/reanahub/reana-db/issues/89:
> (2) opening interactive sessions without necessarily having any workflow yet.

There is now no notion of `workspace` for `InteractiveSessoin`'s, however, the `InteractiveSession`'s implementation depends on it. So far, we have been [depending on the workflow model to retrieve it](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/workflow_run_manager.py#L287:L287), which was ok because the interactive session was also within the workflow model.

As we are evolving to this new idea of separate sessions, I see a need to [extract the workspace concept from within the `Workflow` model](https://github.com/reanahub/reana-db/blob/8203b6afeb90df9d39c13a4c00e71c43cd0e9077/reana_db/models.py#L328) to a new entity. This idea gains more importance once we start thinking about using different workspaces than CephFS.

diegodelemos (commented):
Looks good :) I've added a couple of comments regarding naming and relationships which shouldn't be creating much trouble in your current implementation, once we decide on them, you can create the Alembic recipe.

reana_db/models.py
```diff
@@ -286,6 +286,58 @@ class JobStatus(enum.Enum):
...
+    )
+
+    def __repr__(self):
+        """User Workflow Session string representation."""
+        return "<UserWorkflowSession {} {} {}>".format(
+            self.user_id, self.session_id, self.workflow_id
+        )
+
+
+class InteractiveSessionType(enum.Enum):
```
diegodelemos, line 308: I now realise we have a little problem with this enum.

We already have a [`INTERACTIVE_SESSION_TYPES`](https://github.com/reanahub/reana-commons/blob/e6cdbedb62b21cc20407512e7d0a09eeb28f9481/reana_commons/config.py#L261:L261) config variable in REANA-Commons. We could completely retire it an replace it with this new enum if it were not because of [REANA-Client depending on it](https://github.com/reanahub/reana-client/blob/1e69bc3a8e48769b75b6f8ff267c7974050411ac/reana_client/cli/workflow.py#L1236).

We have two options:
1. Keep the enum, remove `INTERACTIVE_SESSION_TYPES` and therefore "disregard" the [client-side validation in REANA-Client](https://github.com/reanahub/reana-client/blob/1e69bc3a8e48769b75b6f8ff267c7974050411ac/reana_client/cli/workflow.py#L1236) and deal with it in either of the following ways
     1. Letting errors be triggered and rely on the server-side message to tell the user what are the possible values for the interactive sessions
     2. Offer the `InteractiveSessionType` through REST API so we can achieve the client-side validation after a first request retrieving config
2. Keep the `INTERACTIVE_SESSION_TYPES` and all kind of `_TYPES` variables in REANA-Commons because we value the CLI client-side validation in the short term


I think that `1. ii.` is the best option, as it could lead to a future endpoint that can return all of these kind of configs aggregated to both R-Client and R-UI can benefit from it. If we decide on it, for now in R-Client we can live without client-side validation and create and issue to implement the previously mentioned mechanism.

reana_db/models.py
audrium: That's a good solution IMO. @tiborsimko , what do you think?

reana_db/models.py
```diff
@@ -286,22 +286,71 @@ class JobStatus(enum.Enum):
...
+    """Enumeration of interactive session types."""
+
+    jupyter = 0
+
+
+class InteractiveSession(Base, Timestamp):
```
diegodelemos, line 314: We are missing the `InteractiveSessionResource` table ([like `WorkflowResource`](https://github.com/reanahub/reana-db/blob/8203b6afeb90df9d39c13a4c00e71c43cd0e9077/reana_db/models.py#L667)), we will need it to account resource usage in InteractiveSessions (more [here](https://github.com/reanahub/reana-db/pull/92/files#r501686934)).


For now, we should measure just CPU usage, not disk usage since now the workspace is the same as the workflow and it is already accounted (which would change if we go for https://github.com/reanahub/reana-db/pull/90/files#r501548133). Can you remember to open an issue for this, "accounting for `InteractiveSessions` CPU usage" once we merge this?

reana_db/models.py
tiborsimko: Yes, for this sprint and for MVP, it seems perfectly OK indeed. 

(However, we may need to refactor this later, when we shall be addressing user group and access rights. Example: currently, a `Workflow` row has an `owner_id`, but in reality there can be several people associated with a workflow: one main owner, 2-3 collaborators having some R/W rights e.g. to rerun it, a convener having R/O rights to see the results, etc. Some collaborators could run different notebooks on the same workspace, not being owners strictly speaking... These kinds of considerations can naturally wait for later; however it is good to have a future outlook with a flexible "user-workflow-session-role" structure in mind to see where we shall be heading later.)

audrium: @tiborsimko, should we create an issue from your comment? Or this is something to bear in mind?

reana_db/models.py
audrium: I've created an [issue](https://github.com/reanahub/reana-client/issues/430) to address the refactoring in REANA-Client and REANA-Commons, for the rest services refactored the code to use the new enum

mvidalgarcia:
Minor: there's a typo in the last commit title.

reana_db/models.py
```diff
@@ -290,22 +290,66 @@ class JobStatus(enum.Enum):
...
+        default=InteractiveSessionType.jupyter,
+    )
+
+    __table_args__ = (
+        UniqueConstraint("name", "path", name="_interactive_session_uc"),
+        {"schema": "__reana"},
+    )
+
+    def __repr__(self):
+        """Interactive Session string represetantion."""
```
mvidalgarcia, line 336:
```suggestion
        """Interactive Session string representation."""
```

audrium: Nice! You have a good eyes! I've actually copied it from Workflow model, so the issue is there too (and also Users model :D)

diegodelemos: Definitely a sharp eye for detail, thanks Marco 🤗

Issue #88, opened 2020-10-05
models: disk usage to be stored not calculated everytime


diegodelemos:
`get_user_disk_usage` should follow the same approach as `get_user_cpu_usage`, just reading the value from DB and calculate it only if it hasn't been calculated before. The field value should be calculated when:
- Workflow status changes (`workflow_status_change_listener`)
- When https://github.com/reanahub/reana-workflow-controller/issues/333 is executed.

_Originally posted by @diegodelemos in https://github.com/reanahub/reana-db/pull/87#discussion_r497523567_

mvidalgarcia:
When implementing this, https://github.com/reanahub/reana-server/pull/290/commits/fe9ce08922f2daa78d4c2b3254c5bede6d5d3076 should be reverted.

PR #91, opened 2020-10-06
models: overall user quota consumption

diegodelemos:
* Closes reanahub/reana-server#289.

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -137,27 +137,16 @@ class User(Base, Timestamp):
         except Exception:
             return -1
 
-    def get_user_cpu_usage(self):
-        """Aggregate user CPU usage."""
-        try:
-            from .database import Session
+    def get_user_quota_by_type(self, resource_type):
+        """Aggregate user quota usage by resource type."""
+        quota_usage = 0
+        quota_limit = 0
+        for user_resource in self.resources:
+            if user_resource.resource.type_ == resource_type:
+                quota_usage += user_resource.quota_used
+                quota_limit += user_resource.quota_limit
 
-            cpu_resources = [
-                str(res[0])
-                for res in Session.query(Resource.id_)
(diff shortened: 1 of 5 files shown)


reana_db/models.py
```diff
@@ -137,27 +137,16 @@ def get_user_disk_usage(self, block_size=None):
         except Exception:
             return -1
 
-    def get_user_cpu_usage(self):
-        """Aggregate user CPU usage."""
-        try:
-            from .database import Session
+    def get_user_quota_by_type(self, resource_type):
```
mvidalgarcia, line 140: Minor: What do you think about omitting the word `user` from this function signature (and some of the ones above)? In the end, they always called from a user instance.

reana_db/models.py
```diff
@@ -208,6 +197,18 @@ def initialize_user_quota_limits(self):
...
 
+    def get_quota_usage(self):
+        """Get user quota usage information."""
+        return dict(
+            quota=dict(
+                disk=dict(
+                    usage=self.get_user_disk_usage(),
+                    limit=self.get_user_quota_by_type(ResourceType.disk)["limit"],
+                ),
+                cpu=self.get_user_quota_by_type(ResourceType.cpu),
```
mvidalgarcia, line 208: I guess `disk` dict will look something like this eventually, am I right? Once https://github.com/reanahub/reana-db/issues/88 is implemented

diegodelemos: Yes, that was my idea, good that we see it similarly :)

reana_db/models.py
diegodelemos: I think it is a good idea, however, it would imply touching other repos, shall we create an issue for it? Maybe we can add it to this sprint, as a well-isolated task to prettify our current work.

mvidalgarcia (approved):

PR #92, opened 2020-10-07
cli: add command to update users disk quota usage

### LINKED ISSUE #88: models: disk usage to be stored not calculated everytime

mvidalgarcia:
closes reanahub/reana-workflow-controller#333
closes #88

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -16,7 +16,7 @@ from datetime import datetime
 
 from reana_commons.utils import get_disk_usage
 from reana_db.config import DB_SECRET_KEY, DEFAULT_QUOTA_RESOURCES
-from reana_db.utils import build_workspace_path
+from reana_db.utils import build_workspace_path, update_users_disk_quota
 from requests.sessions import session
 from sqlalchemy import (
     BigInteger,
@@ -128,15 +128,6 @@ class User(Base, Timestamp):
         """
         return build_workspace_path(self.id_)
 
-    def get_user_disk_usage(self, block_size=None):
-        """Retrieve user disk usage information."""
-        try:
-            workspace_path = self.get_user_workspace()
-            disk_usage = get_disk_usage(workspace_path, True, block_size)
-            return disk_usage[0]["size"]
-        except Exception:
-            return -1
-
     def get_user_quota_by_type(self, resource_type):
         """Aggregate user quota usage by resource type."""
 
@@ -217,11 +208,7 @@ class User(Base, Timestamp):
     def get_quota_usage(self):
         """Get user quota usage information."""
         return dict(
-            disk=dict(
-                usage=self.get_user_disk_usage(),
-                limit=self.get_user_quota_by_type(ResourceType.disk)["limit"],
-                health=self.get_user_quota_by_type(ResourceType.disk)["health"],
-            ),
+            disk=self.get_user_quota_by_type(ResourceType.disk),
             cpu=self.get_user_quota_by_type(ResourceType.cpu),
         )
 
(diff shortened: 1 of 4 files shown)


reana_db/models.py
```diff
@@ -523,6 +511,7 @@ def workflow_status_change_listener(workflow, new_status, old_status, initiator)
             user_resource_quota.quota_used += cpu_milliseconds
             Session.add(workflow_resource)
             Session.commit()
+        update_users_disk_quota(user=workflow.owner)
```
mvidalgarcia, line 514: This update probably needs to be done in some other places:
- Workflow deletion
- File update (`reana-client update -w roofit myscript.py`)
- Some others?

WDYT?

mvidalgarcia: Another thing, here we're calculating and storing `WorkflowResource` for the CPU but we're no calculating that anywhere else for disk. How should we address this? I think for disk is trickier as it's not necessarily accumulative as a user can add and remove files from the workflow workspace. Probably in this case it needs to be calculated every time? I think the value corresponds to [`get_workspace_disk_usage`](https://github.com/reanahub/reana-db/blob/8203b6afeb90df9d39c13a4c00e71c43cd0e9077/reana_db/models.py#L456-L458) function directly.

diegodelemos: I think:
1. Workflow deletion (`reana-client delete -w w1`) -> rerun the full calculation
2. File upload/deletion (`reana-client rm`/`reana-client upload`) -> we can run the full calculation for now but, this can become dangerous:
    - Doing this calculation on every upload could be a very expensive server-side operation, as we [upload each file on a different HTTP request](https://github.com/reanahub/reana-client/blob/577e5dd52f14a75bed32863ccbd419e6c2ebac0d/reana_client/cli/files.py#L277).
    - We know how big those files are when uploaded, so in theory we can just increment/decrement the disk usage by the number of bytes the files occupy at workflow and user level.

    We could deal with this outside this PR group, as a standalone issue.
3. Interactive session closed (`reana-client close -w w1`) -> run full calculation, no way around it
4. Workflow finishes/fails/stops (addressing the [second question](https://github.com/reanahub/reana-db/pull/92/files#r501577993))-> run a full calculation, and overwrite the value in `WorkflowResource` using `get_workspace_disk_usage` even at the risk of not being fully correct in the overall as a function of time for now (it could be correct if we implement the fix for `2.`). This way we make the workflow list request (mostly coming from the UI) more light (see [here](https://github.com/reanahub/reana-workflow-controller/blob/2173e2c1641e42a22422c4920e520090fe527886/reana_workflow_controller/rest/workflows.py#L239)). `get_workspace_disk_usage` could be checking the DB (`WorkflowResource`) and if there is not calculated disk usage run `du` and update `WorkflowResource`.

Fix for `2.` and going beyond full calculation in `4.` are probably outside of the scope of this PR.

mvidalgarcia: - `1.` Addressed
- RE `3.`, what do you think about creating a status change listener in the [new InteractiveSession model](https://github.com/reanahub/reana-db/pull/90/files) to run `update_users_disk_quota`.
- Maybe we can ticketize the other two based on your comment?

diegodelemos: > RE 3., what do you think about creating a status change listener in the new InteractiveSession model to run update_users_disk_quota.

Tracked through here https://github.com/reanahub/reana-db/pull/90#discussion_r501721868

> Maybe we can ticketize the other two based on your comment?

Let's create the tickets once we merge this? That would be a ticket for `2.` and another for the improvement of `4.`, right? I can do it if you want.

diegodelemos (approved):

PR #96, opened 2020-10-14
models: provide human readable quota usage

diegodelemos:
* Closes reanahub/reana-server#296.

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -11,6 +11,7 @@
 from __future__ import absolute_import
 
 import enum
+import math
 import uuid
 from datetime import datetime
 
@@ -145,14 +146,32 @@ class User(Base, Timestamp):
 
         quota_usage = 0
         quota_limit = 0
+        unit = None
         for user_resource in self.resources:
             if user_resource.resource.type_ == resource_type:
+                # make sure that all resources of the same type use the same units
+                if unit and unit != user_resource.resource.unit:
+                    raise Exception(
+                        "Error while calculating quota usage. Not all "
+                        f"resources of resource type {resource_type} use "
+                        "the same units."
+                    )
+                unit = user_resource.resource.unit
                 quota_usage += user_resource.quota_used
                 quota_limit += user_resource.quota_limit
 
+        quota_human_readable_usage = ResourceUnit.human_readable_unit(unit, quota_usage)
+        quota_human_readable_limit = ResourceUnit.human_readable_unit(unit, quota_limit)
+
         return {
-            "usage": quota_usage,
-            "limit": quota_limit,
+            "usage": {
+                "raw": quota_usage,
+                "human_readable": quota_human_readable_usage,
+            },
+            "limit": {
+                "raw": quota_limit,
+                "human_readable": quota_human_readable_limit,
+            },
             "health": _get_health_status(quota_usage, quota_limit),
         }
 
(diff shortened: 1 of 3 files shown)


reana_db/models.py
```diff
@@ -638,6 +644,51 @@ class ResourceUnit(enum.Enum):
     bytes_ = 0
     milliseconds = 1
 
+    @staticmethod
+    def _human_readable_milliseconds(milliseconds):
+        """Convert bytes usage to human readable string."""
```
mvidalgarcia, line 649:
```suggestion
        """Convert milliseconds usage to human readable string."""
```

tests/test_models.py
```diff
@@ -222,3 +223,32 @@ def test_user_cpu_usage(db, session, new_user, run_workflow):
...
+        (ResourceUnit.bytes_, 1024 * 35, "35 KB"),
+        (ResourceUnit.bytes_, 1024 * 200 + 512, "200.5 KB"),
+        (ResourceUnit.bytes_, 1024 ** 2, "1 MB"),
+        (ResourceUnit.bytes_, 1024 ** 2 + 1024 * 768, "1.75 MB"),
+        (ResourceUnit.bytes_, 1024 ** 3 * 5 + 1024 ** 2 * 100, "5.1 GB"),
+        (ResourceUnit.bytes_, 1024 ** 4 + 1024 ** 3 * 256, "1.25 TB"),
+    ],
+)
+def test_human_readable_unit_values(unit, value, human_readable_string):
+    """Test converting units from canonical values to ."""
```
mvidalgarcia, line 253:
```suggestion
    """Test converting units from canonical values to human-readable string."""
```

reana_db/models.py
```diff
@@ -145,14 +146,19 @@ def _get_health_status(usage, limit):
...
                 quota_usage += user_resource.quota_used
                 quota_limit += user_resource.quota_limit
+                quota_human_readable_usage = ResourceUnit.human_readable_unit(
+                    user_resource.resource.unit, user_resource.quota_used
+                )
 
         return {
             "usage": quota_usage,
             "limit": quota_limit,
+            "human_readable_usage": quota_human_readable_usage,
```
mvidalgarcia, line 161: I believe we also need a `human_readable_limit` to display properly in the clients in a "X KB out of Y MB" way.

reana_db/models.py
```diff
@@ -145,14 +146,19 @@ def _get_health_status(usage, limit):
...
         quota_usage = 0
         quota_limit = 0
+        quota_human_readable_usage = ""
         for user_resource in self.resources:
             if user_resource.resource.type_ == resource_type:
                 quota_usage += user_resource.quota_used
                 quota_limit += user_resource.quota_limit
+                quota_human_readable_usage = ResourceUnit.human_readable_unit(
+                    user_resource.resource.unit, user_resource.quota_used
+                )
```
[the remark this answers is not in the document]
diegodelemos: > Now I wonder what would happen if a user had more than one resource of a specific resource_type, and each of those with a different unit. This accumulation would be wrong, wouldn't it? Because we'd be adding quota_useds and quota_limits no matter the unit. For now it's fine because we only have one unit per resource_type, but this might bring problems if we add other resources/units. Just dumping my thoughts to beware in the future, it doesn't need to be addressed now, as long as we don't increase resource units.

This is true, it is a shortcoming of a very flexible design... maybe we just create an issue to take into account once we add a new resource?

reana_db/models.py
diegodelemos: Makes sense, what do you think of changing the format of the dictionary to something like follows?
```python
        return {
            "usage": {
                "raw": quota_usage,
                "human_readable": quota_human_readable_usage,
            },
            "limit": {
                "raw": quota_limit,
                "human_readable": quota_human_readable_limit,
            },
            "health": _get_health_status(quota_usage, quota_limit),
        }
```

audrium: IMHO nested data structures may be harder to consume in the client side. It could also lead to some problems if there is something wrong with the data and client tries to access nested object (additional checks needed)

mvidalgarcia: With the new [optional chaining ES6 operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) we don't need to do those additional checks anymore, if the key doesn't exist we will get `undefined`.

mvidalgarcia (approved):

PR #99, opened 2020-10-20
models: add default quota limits per resource type

mvidalgarcia:
closes reanahub/reana#425

### DIFF
--- a/reana_db/config.py
+++ b/reana_db/config.py
@@ -51,3 +51,9 @@ DEFAULT_QUOTA_RESOURCES = {
     "disk": "shared storage",
 }
 """Default quota resources to fill Resource table."""
+
+DEFAULT_QUOTA_LIMITS = {
+    "cpu": int(float(os.getenv("REANA_DEFAULT_QUOTA_CPU_LIMIT", 0))),
+    "disk": int(float(os.getenv("REANA_DEFAULT_QUOTA_DISK_LIMIT", 0))),
+}
+"""Default CPU (in milliseconds) and disk (in bytes) quota limits."""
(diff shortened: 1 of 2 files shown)


reana_db/config.py
```diff
@@ -51,3 +51,9 @@
     "disk": "shared storage",
 }
 """Default quota resources to fill Resource table."""
+
+DEFAULT_QUOTA_LIMITS = {
+    "cpu": os.getenv("REANA_DEFAULT_QUOTA_CPU_LIMIT", 0),
+    "disk": os.getenv("REANA_DEFAULT_QUOTA_DISK_LIMIT", 0),
```
diegodelemos, line 57: We should make sure that the value that goes to DB is an integer, however, we will be effectively passing floats because of https://github.com/helm/helm/issues/3001.
```
$ kubectl get pod -o yaml deployment/reana-server
...
    - name: REANA_DEFAULT_QUOTA_CPU_LIMIT                                                                                     
      value: "1.08e+07"
    - name: REANA_DEFAULT_QUOTA_DISK_LIMIT                     
      value: "1.049e+08"    
...
```
To avoid this problem and human errors we should do a explicit cast to `int`.

```suggestion
    "cpu": int(float(os.getenv("REANA_DEFAULT_QUOTA_CPU_LIMIT", 0))),
    "disk": int(float(os.getenv("REANA_DEFAULT_QUOTA_DISK_LIMIT", 0))),
```

mvidalgarcia: Good catch! Tricky one. It's fixed now.

diegodelemos (approved):

Issue #100, opened 2020-10-21
models: update WorkflowResource on workflow deletion


mvidalgarcia:
When a workflow is deleted we're not updating its `WorkflowResource`. E.g. when freeing up disk from the UI, it still shows that it uses some size.

![image](https://user-images.githubusercontent.com/8863238/96731483-971af000-13b7-11eb-8c1c-08958aca3f3b.png)

even though the workspace does not exist anymore:

```console
$ reana-client ls -w hi                  
Something went wrong while retrieving file list for workflow hi:
Workspace does not exist.
$ reana-client du -w hi      
Disk usage could not be retrieved: 
Directory does not exist.
$ reana-client du -w hi -s   
SIZE    NAME
12288   .   
```

PR #101, opened 2020-10-23
models: update workflow disk usage on delete

diegodelemos:
* Refactors `workflow_status_change_listener` to clearly show what
  are the actions to be taken according to the new status.

* Updates workflow disk usage on workflow deletion
  (closes reanahub/reana-db#100).

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -575,43 +575,59 @@ class Workflow(Base, Timestamp):
         current_transition = (self.status, next_status)
         return current_transition in ALLOWED_WORKFLOW_STATUS_TRANSITIONS
 
+    def update_workflow_timestamp(self, new_status):
+        """Update workflow timestamps according to new status."""
+        from .database import Session
+
+        if new_status in [
+            RunStatus.finished,
+            RunStatus.failed,
+        ]:
+            self.run_finished_at = datetime.now()
+        elif new_status in [RunStatus.stopped]:
+            self.run_stopped_at = datetime.now()
+        elif new_status in [RunStatus.running]:
+            self.run_started_at = datetime.now()
+        Session.commit()
+
 
 @event.listens_for(Workflow.status, "set")
 def workflow_status_change_listener(workflow, new_status, old_status, initiator):
     """Workflow status change listener."""
     from .database import Session
 
+    def _update_disk_quota(workflow):
+        update_users_disk_quota(user=workflow.owner)
+        store_workflow_disk_quota(workflow)
+
+    def _update_cpu_quota(workflow):
+        terminated_at = workflow.run_finished_at or workflow.run_stopped_at
+        if workflow.run_started_at and terminated_at:
+            cpu_time = terminated_at - workflow.run_started_at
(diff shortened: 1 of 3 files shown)


reana_db/models.py
```diff
@@ -575,43 +575,56 @@ def can_transition_to(self, next_status):
         current_transition = (self.status, next_status)
         return current_transition in ALLOWED_WORKFLOW_STATUS_TRANSITIONS
 
+    def update_workflow_timestamp(self, new_status):
```
mvidalgarcia, line 578: Out of curiosity, do the changes if this function get commited automatically? Wondering if we need a `Session.commit()` here.

mvidalgarcia (approved):

reana_db/models.py
diegodelemos: Added. It worked because there is an eventual commit in: https://github.com/reanahub/reana-db/blob/6c5ebc2a69b54818652957e6867a66f022eccf04/reana_db/models.py#L548-L571

After updating the status.

reana_db/utils.py
```diff
@@ -256,7 +258,7 @@ def store_workflow_disk_quota(workflow):
         disk_bytes = 0
     if workflow_resource:
         workflow_resource.quantity_used = disk_bytes
-    else:
+    elif inspect(workflow).persistent:
```
mvidalgarcia, line 261: Question: what is the root problem that leads us to do this change? Is it that the workflow is not yet committed and we do `Session.add(workflow_resource)` and the workflow does not exist yet on DB?
It looks a bit hacky IMO, but if there's no other way...

BTW, now that there's an `elif`, if none of these checks pass, we'd be doing a `Session.commit()` in any case, is that correct?

diegodelemos: > Question: what is the root problem that leads us to do this change? Is it that the workflow is not yet committed and we do Session.add(workflow_resource) and the workflow does not exist yet on DB?
It looks a bit hacky IMO, but if there's no other way...

I completely agree, this is a hack. The root cause is that, using [the status change listener approach](https://github.com/reanahub/reana-db/blob/c03aa75fed2b0c5d3aa4cc3555bf72ac3a3cf6b0/reana_db/models.py#L579-L580) might not be the best approach to handle these situations because the status changes [when the SQLAlchemy object that represents the workflow is created in memory (in the constructor)](https://github.com/reanahub/reana-db/blob/c03aa75fed2b0c5d3aa4cc3555bf72ac3a3cf6b0/reana_db/models.py#L432), hence, triggering the listener when the workflow is not yet in DB. It happened that we discovered it now because we try to make reference to the workflow table for the first time inside the listener logic. The options I was left with right now were: (1) refactoring or (2) adding this patch to get things going for now.

My fault for not correctly stating (1) and (2) in the first place and justifying that (2) is chosen because of time/blocking other issues.

@mvidalgarcia what do you think about creating an issue once this is merged so we can reference to the problem in the code?

> BTW, now that there's an elif, if none of these checks pass, we'd be doing a Session.commit() in any case, is that correct?

Thanks, good catch, I am amending it.

PR #105, opened 2020-10-28
models: add function to check if quota is exceeded

mvidalgarcia:
closes reanahub/reana-server#282

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -242,6 +242,12 @@ class User(Base, Timestamp, QuotaBase):
                 )
             )
 
+    def has_exceeded_quota(self):
+        """Get whether user has exceeded the quota of any resource."""
+        return any(
+            r.quota_limit != 0 and r.quota_used >= r.quota_limit for r in self.resources
+        )
+
     def __repr__(self):
         """User string representation."""
         return "<User %r>" % self.id_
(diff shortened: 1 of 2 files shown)


reana_db/models.py
```diff
@@ -236,6 +236,12 @@ def get_quota_usage(self):
             cpu=self.get_user_quota_by_type(ResourceType.cpu),
         )
 
+    def has_exceeded_quota(self):
```
mvidalgarcia, line 239: I'm wondering if there should be a check per resource type as there might be some actions that the user can perform even when only the CPU quota is exceeded (create workflows, upload files, etc.) and the other way around (starting workflows).

mvidalgarcia: Decided offline to remain conservative and prevent the mentioned user actions when the quota of **any** user resource is exceeded.

reana_db/models.py
```diff
@@ -242,13 +242,19 @@ def initialize_user_quota_limits(self):
                 )
             )
 
+    def has_exceeded_quota(self):
+        """Get whether user has exceeded the quota of any resource."""
+        return any(
+            r.quota_limit != 0 and r.quota_used >= r.quota_limit for r in self.resources
+        )
+
+>>>>>>> models: add function to check if quota is exceeded
```
alintulu, line 251: Possible left over from a merge conflict?

mvidalgarcia: 🤦🏻‍♂️

diegodelemos (approved):

PR #111, opened 2020-11-05
models: use powers of 1024 to display size units

mvidalgarcia:
E.g. "KiB", "MiB", "GiB", etc.

closes reanahub/reana#437

---

```console
$ reana-client quota-show --resource disk               
222.55 KiB out of 8.58 MiB used (3%)
```

image

image

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changes
 Version master (UNRELEASED)
 ---------------------------
 
-- Adds disk usage retreival methods to user and workflow models.
+- Adds new disk usage retrieval methods using canonical (bytes) and human-readable (KiB) units. (``User``, ``Workflow``)
 
 Version 0.7.0 (2020-10-20)
 --------------------------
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -4,7 +4,8 @@ Changes
 Version master (UNRELEASED)
 ---------------------------
 
-- Adds disk usage retreival methods to user and workflow models.
+- Adds disk usage retrieval methods to user and workflow models.
+- Changes human-readable size units to display them in powers of 1024.
```
tiborsimko, line 8: Since there were no quota units in 0.7.0, there is no change when compared to the previous release... So instead of Changes, it should say Adds.  But you can simply join with the previous item, perhaps something like:

- Adds new disk usage retrieval methods using canonical (bytes) and human-readable (KiB) units. (``User``, ``Workflow``)

tiborsimko (approved):

PR #119, opened 2021-02-02
release: 0.7.1

tiborsimko:


setup.py
```diff
@@ -20,6 +20,7 @@
 
 tests_require = [
     "pytest-reana>=0.7.0a1,<0.8.0",
+    'google-auth==1.24.0 ; python_version=="2.7"',
```
tiborsimko, line 23: FWIW this fix should rather belong to `r-commons[kubernetes]`, but since we released that one already, and since we don't use py27 except for the client, perhaps leaving it here is OK for now...

diegodelemos (approved):

PR #124, opened 2021-04-26
models: add start and finish times for jobs

diegodelemos:
Closes reanahub/reana-client#497.

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -663,8 +663,27 @@ class Job(Base, Timestamp):
     env_vars = Column(JSONType)
     deleted = Column(Boolean)
     logs = Column(String, nullable=True)
+    started_at = Column(DateTime)
+    finished_at = Column(DateTime)
     prettified_cmd = Column(JSONType)
     job_name = Column(Text)
+
+
+@event.listens_for(Job.status, "set")
+def job_status_change_listener(job, new_status, old_status, initiator):
+    """Job status change listener."""
+    if new_status != old_status:
+        from .database import Session
+
+        if new_status in [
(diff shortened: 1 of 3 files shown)


ParthS007 (approved):
Thanks for the PR. Tested locally, working fine 🚀

reana_db/models.py
```diff
@@ -656,10 +656,28 @@ class Job(Base, Timestamp):
...
+    if new_status != old_status:
+        if new_status in [
+            JobStatus.finished,
+            JobStatus.failed,
+        ]:
+            job.finished_at = datetime.now()
+        elif new_status in [JobStatus.running]:
+            job.started_at = datetime.now()
+
+    Session.commit()
```
mvidalgarcia, line 678: Should we commit in the _very rare_ case that `new_status` == `old_status`? Maybe this commit should only happen when they're different, what do you think?

mvidalgarcia (approved):
Haven't tested it. Code looks good 👍

Just a minor comment.

reana_db/models.py
```diff
@@ -663,10 +663,29 @@ class Job(Base, Timestamp):
     env_vars = Column(JSONType)
     deleted = Column(Boolean)
     logs = Column(String, nullable=True)
+    started_at = Column(DateTime)
+    finished_at = Column(DateTime)
```
tiborsimko, line 667: Do we have Alembic upgrade recipe for this change? Or shall we compile them afterwards before releasing?

mvidalgarcia: Definitely needed.

PR #128, opened 2021-05-07
alembic: add revision for job started/finished times

mvidalgarcia:
closes #127

To test:
```console
$ kubectl exec -ti deployment/reana-db -- psql -U reana
psql (9.6.2)
Type "help" for help.

reana=# \d __reana.job
                     Table "__reana.job"
       Column       |            Type             | Modifiers 
--------------------+-----------------------------+-----------
...
 finished_at        | timestamp without time zone | 
 started_at         | timestamp without time zone | 
...
reana=# \q

$ kubectl exec deployment/reana-server reana-db alembic history
ad93dae04483 -> 4801b98f6408 (head), Job started and finished times.
c912d4f1e1cc -> ad93dae04483, Interactive sessions.
<base> -> c912d4f1e1cc, Quota tables.

$ kubectl exec deployment/reana-server reana-db alembic current
4801b98f6408 (head)

$ kubectl exec deployment/reana-server reana-db alembic downgrade ad93dae04483
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running downgrade 4801b98f6408 -> ad93dae04483, Job started and finished times.

$ kubectl exec -ti deployment/reana-db -- psql -U reana
psql (9.6.2)
Type "help" for help.

reana=# \d __reana.job
                     Table "__reana.job"
       Column       |            Type             | Modifiers 
--------------------+-----------------------------+-----------
...
👀  NO FINISHED_AT AND STARTED_AT!
...
reana=# \q

$ kubectl exec deployment/reana-server reana-db alembic upgrade
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade ad93dae04483 -> 4801b98f6408, Job started and finished times.

$ kubectl exec -ti deployment/reana-db -- psql -U reana
psql (9.6.2)
Type "help" for help.

reana=# \d __reana.job
                     Table "__reana.job"
       Column       |            Type             | Modifiers 
--------------------+-----------------------------+-----------
...
 finished_at        | timestamp without time zone | 
 started_at         | timestamp without time zone | 
...
reana=# \q
```

PR #130, opened 2021-05-20
models: store workflow complexity

audrium:
closes https://github.com/reanahub/reana-server/issues/358

#### Test alembic revision:

```console
$ kubectl exec deployment/reana-server reana-db alembic history
4801b98f6408 -> f84e17bd6b18 (head), Workflow complexity.
ad93dae04483 -> 4801b98f6408, Job started and finished times.
c912d4f1e1cc -> ad93dae04483, Interactive sessions.
<base> -> c912d4f1e1cc, Quota tables.

$ kubectl exec deployment/reana-server reana-db alembic current
f84e17bd6b18 (head)

$ kubectl exec deployment/reana-server reana-db alembic downgrade 4801b98f6408
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running downgrade f84e17bd6b18 -> 4801b98f6408, Workflow complexity.

$ kubectl exec deployment/reana-server reana-db alembic upgrade
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
f84e17bd6b18 (head)
```
You can check the status of PostgreSQL using `psq`l as indicated [here](https://github.com/reanahub/reana-db/pull/128).

⚠️ It works with empty DB (no workflows created). As soon as you create a workflow (`reana-client run`), it gets stuck on downgrade.

`ALTER TABLE __reana.workflow DROP COLUMN complexity;` or `ALTER TABLE __reana.workflow ADD COLUMN mycomplexity VARCHAR NOT NULL DEFAULT '[]';` also gets stuck via `psql` so it seems there is a PostgreSQL problem.

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -38,6 +38,7 @@ from sqlalchemy.orm import relationship
 from sqlalchemy_utils import EncryptedType, JSONType, UUIDType
 from sqlalchemy_utils.models import Timestamp
 from sqlalchemy_utils.types.encrypted.encrypted_type import AesEngine
+from sqlalchemy.dialects.postgresql import ARRAY
 
 from reana_db.config import DB_SECRET_KEY, DEFAULT_QUOTA_LIMITS, DEFAULT_QUOTA_RESOURCES
 from reana_db.utils import (
@@ -416,6 +417,7 @@ class Workflow(Base, Timestamp, QuotaBase):
     reana_specification = Column(JSONType)
     input_parameters = Column(JSONType)
     operational_options = Column(JSONType)
+    complexity = Column(ARRAY(BigInteger(), dimensions=2), default=[])
     type_ = Column(String(30))
     logs = Column(String)
     run_started_at = Column(DateTime)
@@ -462,6 +464,7 @@ class Workflow(Base, Timestamp, QuotaBase):
         input_parameters={},
         operational_options={},
         status=RunStatus.created,
+        complexity=[],
         git_ref="",
         git_repo=None,
         git_provider=None,
@@ -477,6 +480,7 @@ class Workflow(Base, Timestamp, QuotaBase):
         self.reana_specification = reana_specification
         self.input_parameters = input_parameters
         self.operational_options = operational_options
+        self.complexity = complexity
         self.type_ = type_
         self.logs = logs or ""
         self.git_ref = git_ref
--- a/reana_db/alembic/versions/20210607_1217_f84e17bd6b18_workflow_complexity.py
+++ b/reana_db/alembic/versions/20210607_1217_f84e17bd6b18_workflow_complexity.py
@@ -1,0 +1,35 @@
+"""Workflow complexity.
+
+Revision ID: f84e17bd6b18
+Revises: 4801b98f6408
+Create Date: 2021-06-07 12:17:47.218408
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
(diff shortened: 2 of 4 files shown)


mvidalgarcia:
Just a reminder that it needs an alembic recipe. Check `reana-db` CLI.

mvidalgarcia:
Needs rebase as well.

reana_db/models.py
```diff
@@ -416,6 +417,8 @@ class Workflow(Base, Timestamp, QuotaBase):
     reana_specification = Column(JSONType)
     input_parameters = Column(JSONType)
     operational_options = Column(JSONType)
+    complexity = Column(String, default="[]")
+    workflow_priority = Column(Integer, default=0)
```
mvidalgarcia, line 421: It belongs to the `Workflow` model, so we can call it just `priority`.

mvidalgarcia (commented):
It needs an alembic recipe. Check `reana-db` CLI.

reana_db/models.py
mvidalgarcia: Actually, I think we don't even need a `priority` column as it can be easily computed on the fly combining complexity and overload. Same as I do in [my PR](https://github.com/reanahub/reana-db/pull/131/files#diff-9c9ecb5b02eee5df82a9c1138c56e62181e9d8d56152d45c09f7fe0a693ce980R531).

I can do it in my task if needed, together with the alembic recipe if we want to merge fast.

audrium: You are right! Since we moved the logic to r-server from r-w-c and set priority in the message body, I guess there is no point to store it in database anymore. I've removed it and left only the complexity, since it may be useful in the future.

audrium: >I can do it in my task if needed, together with the alembic recipe if we want to merge fast.

I was thinking exactly the same thing! Since you are also touching the database, probably it would be easier to do it once

mvidalgarcia: Ok, noted in [my PR todo list](https://github.com/reanahub/reana-db/pull/131).

reana_db/models.py
```diff
@@ -416,6 +416,7 @@ class Workflow(Base, Timestamp, QuotaBase):
     reana_specification = Column(JSONType)
     input_parameters = Column(JSONType)
     operational_options = Column(JSONType)
+    complexity = Column(String, default="[]")
```
mvidalgarcia, line 419: Any particular reason why we store complexity as `String`? We could use SQLAlchemy [`ARRAY`](https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.ARRAY), or even JSON.

audrium: I was a bit hesitating to try it since `ARRAY` is only supported by `PostgreSQL`. However just tried it worked as expected

mvidalgarcia (commented):
A couple of things:
- We have to fix the py2 tests. Something similar to https://github.com/reanahub/reana-client/pull/519/files likely.
- We need to create a release commit.

audrium:
>     We have to fix the py2 tests. Something similar to https://github.com/reanahub/reana-client/pull/519/files likely.

Should be fine after merging [pytest-reana PR](https://github.com/reanahub/pytest-reana/pull/83)

mvidalgarcia (approved):
Tests working locally both in py2 and py3 when installing `pytest-reana` PR ✅ 

Pending: Make alembic recipe downgrade work properly when DB has workflows.

PR #131, opened 2021-05-21
models: add workflow priority property

mvidalgarcia:
Closes reanahub/reana-server#356

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -12,11 +12,16 @@ from __future__ import absolute_import
 
 import enum
 import math
+import logging
 import uuid
 from datetime import datetime
-import logging
+from functools import reduce
 
-from reana_commons.config import REANA_RUNTIME_KUBERNETES_KEEP_ALIVE_JOBS_WITH_STATUSES
+from reana_commons.config import (
+    MQ_MAX_PRIORITY,
+    REANA_MAX_CONCURRENT_BATCH_WORKFLOWS,
+    REANA_RUNTIME_KUBERNETES_KEEP_ALIVE_JOBS_WITH_STATUSES,
+)
 from reana_commons.utils import get_disk_usage
 from sqlalchemy import (
     BigInteger,
@@ -31,6 +36,7 @@ from sqlalchemy import (
     UniqueConstraint,
     event,
     func,
+    or_,
 )
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.ext.hybrid import hybrid_property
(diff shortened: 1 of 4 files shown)


reana_db/models.py
```diff
@@ -252,6 +258,18 @@ def has_exceeded_quota(self):
...
+    def get_workflow_overload_priority(self):
+        """Get priority factor based on the number of current workflows ``running``."""
+        max_concurrent_workflows = REANA_MAX_CONCURRENT_BATCH_WORKFLOWS
+        running_count = self.workflows.filter(
+            or_(
+                Workflow.status == RunStatus.pending,
+                Workflow.status == RunStatus.running,
+            )
+        ).count()
+        priority = round(1 - running_count / max_concurrent_workflows, 2)
```
audrium, line 270: If  `running_count` will be bigger then `max_concurrent_workflows` the overload priority will start increasing the workflow priority (in `get_priority()`) instead of decreasing it. That's why I've added additional checks over [here](https://github.com/reanahub/reana-server/blob/master/reana_server/complexity.py#L38)

mvidalgarcia: If I understand it correctly, `running_count` can't be greater than `max_concurrent_workflows` due to [this check](https://github.com/reanahub/reana-server/blob/9bddf50060929313925ae82fa5a377697e842504/reana_server/scheduler.py#L68-L75), but I might be wrong.

tests/test_models.py
```diff
@@ -302,3 +302,81 @@ def test_should_cleanup_job(
...
+    workflow_name = "test-workflow"
+    workflow = Workflow(
+        id_=str(uuid4()),
+        name=workflow_name,
+        owner_id=new_user.id_,
+        reana_specification=[],
+        type_="serial",
+        logs="",
+        status=RunStatus.running,
+    )
```
audrium, line 379: Minor: I see that there is a `run_workflow()` fixture defined in `reana-db`, have you tried to use it?

reana_db/models.py
audrium: I guess you are right, but in case someone will modify this logic in the future we could easily exceed `MQ_MAX_PRIORITY` and break the scheduler. So I wonder if we should introduce some check inside `get_priority()` just to be extra safe?

reana_db/models.py
```diff
@@ -252,6 +258,20 @@ def has_exceeded_quota(self):
             r.quota_limit != 0 and r.quota_used >= r.quota_limit for r in self.resources
         )
 
+    def get_workflow_overload_priority(self):
+        """Get priority factor based on the number of current workflows ``running``."""
+        max_concurrent_workflows = REANA_MAX_CONCURRENT_BATCH_WORKFLOWS
+        running_count = self.workflows.filter(
```
audrium, line 264: Question: shouldn't we filter workflows by user? If I'm not mistaken now if user A submits 10 workflows and after user B submits his first workflow, the priority of that workflow will still be penalized?

tests/test_models.py
mvidalgarcia: Good point, I hadn't seen that fixture. I've changed it together with `test_get_workflow_overload_priority` and it simplifies the tests a lot, thanks!

reana_db/models.py
mvidalgarcia: `self.workflows` is a 1:n (User:Workflow) relationship that already filters the workflows whose owner is `self`.

audrium: I had a suspicion that this is the case, thanks for clarifying!

audrium (approved):
LGTM! :rocket:  Should we update `CHANGES.RST`?

Issue #94, opened 2020-10-09
models: extract the workspace concept from within the Workflow model 


audrium:
Addressing the following line in https://github.com/reanahub/reana-db/issues/89:
> (2) opening interactive sessions without necessarily having any workflow yet.

There is now no notion of `workspace` for `InteractiveSession`'s, however, the `InteractiveSession`'s implementation depends on it. So far, we have been [depending on the workflow model to retrieve it](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/workflow_run_manager.py#L287:L287), which was ok because the interactive session was also within the workflow model.

As we are evolving to this new idea of separate sessions, I see a need to [extract the workspace concept from within the `Workflow` model](https://github.com/reanahub/reana-db/blob/8203b6afeb90df9d39c13a4c00e71c43cd0e9077/reana_db/models.py#L328) to a new entity. This idea gains more importance once we start thinking about using different workspaces than CephFS.

_Originally posted by @diegodelemos in https://github.com/reanahub/reana-db/pull/90#discussion_r501548133_

PR #139, opened 2021-07-22
models: modify workspace_path to store full path

marcdiazsan:
Quit the dependency of upload-file,remove-file, list-files,mv-files  and download, relying directly on the full-path stored in the database
    
Closes reanahub/reana-db#94

### DIFF
--- a/reana_db/utils.py
+++ b/reana_db/utils.py
@@ -13,16 +13,25 @@ from uuid import UUID
 from sqlalchemy import inspect
 
 
-def build_workspace_path(user_id, workflow_id=None):
+def build_workspace_path(user_id, workflow_id=None, workspace_root_path=None):
     """Build user's workspace relative path.
 
     :param user_id: Owner of the workspace.
     :param workflow_id: Optional parameter, if provided gives the path to the
         workflow workspace instead of just the path to the user workspace.
-    :return: String that represents the workspace relative path.
-        i.e. users/0000/workflows/0034
+    :param workspace_root_path: Optional parameter, if provided changes the
+        root path under which the workflow workspaces are stored.
+    :return: String that represents the workspace absolute path.
+        i.e. /var/reana/users/0000/workflows/0034
     """
-    workspace_path = os.path.join("users", str(user_id), "workflows")
+    from reana_commons.config import SHARED_VOLUME_PATH
+
+    if workspace_root_path:
+        workspace_path = workspace_root_path
+    else:
+        workspace_path = os.path.join(
+            SHARED_VOLUME_PATH, "users", str(user_id), "workflows"
+        )
     if workflow_id:
         workspace_path = os.path.join(workspace_path, str(workflow_id))
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -10,9 +10,15 @@
 
 from __future__ import absolute_import, print_function
 
+import os
+
+from reana_commons.config import SHARED_VOLUME_PATH
+
 
 def test_build_workspace_path():
     """Tests for build_workspace_path()."""
     from reana_db.utils import build_workspace_path
 
-    assert build_workspace_path(0) == "users/0/workflows"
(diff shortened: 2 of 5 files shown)


mvidalgarcia:
It looks that something went wrong when creating this PR, as the diff is `+0 −0`.

audrium:
You can also mention in the PR description and the commit message Closes #94 to automatically link this PR with the issue. It's only partially linked with "Adresses", not being displayed in the "Linked pull requests" box of the issue

audrium:
You could also change the PR title to something more meaningful, e.g. "models: modify workspace_path to store full path"

reana_db/models.py
```diff
@@ -513,9 +513,7 @@ def __init__(
         self.git_provider = git_provider
         self.restart = restart
         self._run_number = self.assign_run_number(run_number)
-        self.workspace_path = workspace_path or build_workspace_path(
-            self.owner_id, self.id_
-        )
+        self.workspace_path = workspace_path
```
audrium, line 516: I think we want to keep the old code here, since we need to call `build_workspace_path()` once the workflow initializes

audrium (commented):
Please have a look into my previous comments above.
Also note that tests are not passing, there is one called `test_build_workspace_path()` which needs to be fixed

tests/test_utils.py
```diff
@@ -9,10 +9,14 @@
 """REANA-DB utils tests."""
 
 from __future__ import absolute_import, print_function
+from reana_commons.config import SHARED_VOLUME_PATH
+import os
```
audrium, line 13: We have a convention to separate future statements, python modules and custom modules with an empty line, so in this case it could be:
```python

import os

from reana_commons.config import SHARED_VOLUME_PATH
```

audrium:
LGTM! I think what is left is to squash all the commits to a single one and to rebase against the latest master

reana_db/utils.py
```diff
@@ -22,7 +22,14 @@ def build_workspace_path(user_id, workflow_id=None):
     :return: String that represents the workspace relative path.
         i.e. users/0000/workflows/0034
     """
-    workspace_path = os.path.join("users", str(user_id), "workflows")
+    from reana_commons.config import SHARED_VOLUME_PATH
+
+    if shared_volume:
+        workspace_path = os.path.join(shared_volume, "users", str(user_id), "workflows")
```
audrium, line 28: As we discussed, we don't want to append "/users/{user_id}/workflows/" to the user specified custom workspace path. So I guess in this line we could only set `workspace_path = shared_volume`. But then probably the `shared_volume` variable name is not entirely correct, maybe `workspace_root_path` could be used as in `reana-workflow-controller`?

Also this function misses param docstring in line ~20

marcdiazsan: In the original case when I have `var/reana` should "/users/{user_id}/workflows/" also be deleted?

audrium (approved):
LGTM! :rocket:

reana_db/utils.py
```diff
@@ -13,16 +13,25 @@
...
+def build_workspace_path(user_id, workflow_id=None, workspace_root_path=None):
     """Build user's workspace relative path.
 
     :param user_id: Owner of the workspace.
     :param workflow_id: Optional parameter, if provided gives the path to the
         workflow workspace instead of just the path to the user workspace.
-    :return: String that represents the workspace relative path.
-        i.e. users/0000/workflows/0034
+    :param workspace_root_path: Optional parameter, if provided changes the
+        default directory where the workflow is stored.
```
tiborsimko, line 23: Please amend doctring based on changes in pytest-reana.

Please add yourself to the AUTHORS list.

Please add a release commit on top.

tiborsimko (approved):

tiborsimko:
One more thing: shall we increment `pytest-reana` version that this commit depends on?

```console
$ rg pytest-reana                
setup.py
22:    "pytest-reana>=0.8.0a5,<0.9.0",
```

Issue #141, opened 2021-08-09
alembic: support migration to absolute worspaces


tiborsimko:
After #139 we may be in a situation when a cluster instance has old workflow entries stored with workspace paths stored in relative manner, whist new workflow workspace paths in absolute manner.  This may result in consistence issues.

Two possibilities:

- amend codebase to accept workspace path to be either absolute (when starting with `/`) or relative (for old content)
- write an alembic migration recipe, so that when the cluster will get upgraded by the admins, the DB table workspace path values will also get upgraded from relative to absolute paths.

I think the latter is much more advantageous than the former, but we have to beware about possible DB issues while updating content, so be careful about rollbacks etc.

tiborsimko:
I.e. before, we had:

```console
kubectl exec -i -t deployment/reana-db -- psql -U reana -c 'SELECT workspace_path FROM __reana.workflow'
                                      workspace_path                                       
-------------------------------------------------------------------------------------------
 users/00000000-0000-0000-0000-000000000000/workflows/f2f66cc1-5bc1-4bac-a625-a1747ef87355
 users/00000000-0000-0000-0000-000000000000/workflows/22d5b0b6-9ab4-4957-ad6b-f0d8abde8ad0
 users/00000000-0000-0000-0000-000000000000/workflows/26277180-7749-48dc-aaed-a97634bb63f7
(3 rows)
```

Now, we have something like:

```console$ kubectl exec -i -t deployment/reana-db -- psql -U reana -c 'SELECT workspace_path FROM __reana.workflow'
                 workspace_path                  
-------------------------------------------------
 /tmp/reana/2a27c99c-062d-4918-b992-9dc6277ebb39
 /tmp/reana/92c585c4-b900-431b-b532-67933f2cd0ef
 /tmp/reana/1855f88d-b346-41cf-af50-8478dc6963d4
 /tmp/reana/cbcb2f50-9811-4397-9f49-bd66a27892d1
 /tmp/reana/6fe8691e-e665-42d7-968d-ffc5a24394f6
(5 rows)
```

For testing the migration script, if there are any rows not starting with `/`, you should update the values and put `/var/reana/` in front.

To test it, you can amend some rows via `psql` command like in the following example:

```console
$ kubectl exec -i -t deployment/reana-db -- psql -U reana -c 'SELECT workspace_path FROM __reana.workflow'
                 workspace_path                  
-------------------------------------------------
[...]

PR #142, opened 2021-08-16
alembic: upgrade db to store full workspace path adding the shared volume path

marcdiazsan:
Closes reanahub/reana-db#141


reana_db/alembic/versions/20210816_0752_6568d7cb6710_full_workflow_workspace.py
```diff
@@ -0,0 +1,60 @@
+"""full workflow workspace.
```
tiborsimko, line 1: "Storing full workspace path." might be better description.

reana_db/alembic/versions/20210816_0752_6568d7cb6710_full_workflow_workspace.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def upgrade():
+    """Upgrade to 6568d7cb6710 revision."""
+    connection = op.get_bind()
+    connection.execute(
+        workflow.update()
+        .where(sa.not_(workflow.c.workspace_path.like("/%")))
+        .values(
+            {"workspace_path": SHARED_VOLUME_PATH + "/" + workflow.c.workspace_path}
```
tiborsimko, line 40: Instead of `foo + "/" + bar`. you may write `os.path.join(foo, bar)` which may be more readable.  (And it would ready to work on some exotic systems where the directory separator and/or joining process may differ!)

reana_db/alembic/versions/20210816_0752_6568d7cb6710_full_workflow_workspace.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def downgrade():
+    """Downgrade to f84e17bd6b18 revision."""
+    connection = op.get_bind()
+    connection.execute(
+        workflow.update().values(
+            {
+                "workspace_path": "users"
+                + "/"
```
tiborsimko, line 52: Ditto here, `oo.path.join(dir1, dir2, dir3...)` would be nicer.

reana_db/alembic/versions/20210816_0752_6568d7cb6710_full_workflow_workspace.py
marcdiazsan: Hmm, there is some issue there, os.path.join doesn't accept `workflow.c.workspace_path`, as it is from a `ColumnClause` type, not a str or similar. Perhaps I could use `foo+os.path.sep+bar`?

tiborsimko: Have you tried explicit string cast? Something like: 

```python
os.path.join(SHARED_VOLUME_PATH, str(workflow.c.workspace_path))
```

marcdiazsan: Yes, It stores the name of the column but not the value I need, something like `/var/reana/__reana.workkflow.c.workspace_path`

tiborsimko: OK, so let's go with the original format then (perhaps with `os.sep` if you prefer, but not necessary...)

PR #143, opened 2021-08-24
utils: use default workspace path instead of shared workspace path  when the workspace root path is not declared

marcdiazsan:
Closes reanahub/reana#532


audrium (approved):
LGTM!

audrium (approved):

PR #146, opened 2021-10-14
utils: fix workspace root path usage in shared volume

audrium:
There is a problem where user can specify `workspace root path` to be located inside `reana shared path` (`/var/reana`) or even some specific folder like (`/var/reana/db`). In this case current logic would simply append workflow id and happily use the new workspace directory. 

This PR fixes this problem by checking if `SHARED_VOLUME_PATH` is used inside user's specified `workspace root path` and in that case always use the default shared path directory for workspaces (`/var/reana/users/{user_id}/workflows/{workflow_id}`).

To test create some workflows by specifying different workspace root path's in `reana.yaml`:
```yaml
workspace:
  root_path: /var/reana/db
```
and inspect the workspace paths with:
```
$ kubectl exec -i -t deployment/reana-db -- psql -U reana -c 'SELECT workspace_path FROM __reana.workflow'
                                            workspace_path
------------------------------------------------------------------------------------------------------
 /var/reana/users/00000000-0000-0000-0000-000000000000/workflows/5d14aa3c-3471-4118-a55b-e9dd9d06cdd8
 /var/reana/users/00000000-0000-0000-0000-000000000000/workflows/6be61a50-88ad-4f2a-8a68-a8e6a28be2bf
 /eos/myanalysis/ebd3b0ff-0da8-4949-b5d3-8cf6fee23a31
 /eos/myanalysis/2b998f98-63cf-4063-a528-a1161e7d6207
```


tests/test_utils.py
```diff
@@ -22,3 +22,20 @@ def test_build_workspace_path():
...
+        == "/eos/myanalisis/1"
+    )
+
+    assert build_workspace_path(
+        0, workflow_id=1, workspace_root_path=SHARED_VOLUME_PATH
+    ) == os.path.join(SHARED_VOLUME_PATH, "users/0/workflows/1")
+
+    assert build_workspace_path(
+        0, workflow_id=1, workspace_root_path=SHARED_VOLUME_PATH + "/db"
+    ) == os.path.join(SHARED_VOLUME_PATH, "users/0/workflows/1")
```
mvidalgarcia, line 41: I think this might be a nice use case of [parametrize](https://docs.pytest.org/en/6.2.x/parametrize.html), as we're calling the same function always.

tests/test_utils.py
```diff
@@ -22,3 +22,20 @@ def test_build_workspace_path():
     assert build_workspace_path(0) == os.path.join(
         SHARED_VOLUME_PATH, "users/0/workflows"
     )
+
+    assert build_workspace_path(0, workflow_id=1) == os.path.join(
```
mvidalgarcia, line 26: What about calling this function passing the `user_id` arg name to improve readability?

```
build_workspace_path(user_id=0, workflow_id=1)
```

PR #150, opened 2021-10-21
add env variable to decide which quotas to update

### LINKED ISSUE #149: profile workflow pod termination tasks
During scalability tests, it was observed that pods were a long time in "Terminating" status.

This is probably because of disk quota calculations and updating.

Let us profile the pod termination tasks and measure what takes time.

In order to do this, it'll be good to introduce a new environment variable, such as `REANA_WORKFLOW_TERMINATION_QUOTA_UPDATE_POLICY`, so that we can influence these during scalability tests easily.

E.g. policy=cpu,hdd would mean that both quotas are updated, as is current default.

E.g. policy=hdd would update only HDD quota. (this is presumably slow)

E.g. policy=cpu would update only CPU quota numbers. (this is presumably fast) 

E.g. policy=none would mean that no quota is updated, so that pods terminate as fast as possible. And the quota update would have to be relegated to a nightly cron task, for example.

Note that profiling could advise about any slow-downs, so that the fine granularity like that might not be needed. If the bottleneck is the HDD calculation, then a flag to switch it off and rely on a nightly cron might be enough.

VMois:
Closes #149 

Addresses https://github.com/reanahub/reana/issues/562

### To test

Add `REANA_WORKFLOW_QUOTA_UPDATE_POLICY` to `r-workflow-controller` component in `helm/reana/values.yaml` and redeploy the cluster. Depending on the value (`cpu,disk`, `cpu`, etc.) some quotas will be calculated and some not. You can check in debug mode if quotas are recalculated by checking if execution goes to `update_disk_quota`, etc. functions.

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -46,7 +46,13 @@ from sqlalchemy_utils.models import Timestamp
 from sqlalchemy_utils.types.encrypted.encrypted_type import AesEngine
 from sqlalchemy.dialects.postgresql import ARRAY
 
-from reana_db.config import DB_SECRET_KEY, DEFAULT_QUOTA_LIMITS, DEFAULT_QUOTA_RESOURCES
+from reana_db.config import (
+    DB_SECRET_KEY,
+    DEFAULT_QUOTA_LIMITS,
+    DEFAULT_QUOTA_RESOURCES,
+    WORKFLOW_TERMINATION_QUOTA_UPDATE_POLICY,
+    QuotaResourceType,
+)
 from reana_db.utils import (
     build_workspace_path,
     store_workflow_disk_quota,
@@ -684,10 +690,15 @@ def workflow_status_change_listener(workflow, new_status, old_status, initiator)
     from .database import Session
 
     def _update_disk_quota(workflow):
+        if QuotaResourceType.disk not in WORKFLOW_TERMINATION_QUOTA_UPDATE_POLICY:
+            return
         update_users_disk_quota(user=workflow.owner)
         store_workflow_disk_quota(workflow)
 
     def _update_cpu_quota(workflow):
+        if QuotaResourceType.cpu not in WORKFLOW_TERMINATION_QUOTA_UPDATE_POLICY:
+            return
+
         terminated_at = workflow.run_finished_at or workflow.run_stopped_at
         if workflow.run_started_at and terminated_at:
             cpu_time = terminated_at - workflow.run_started_at
--- a/reana_db/config.py
+++ b/reana_db/config.py
@@ -9,6 +9,7 @@
 """REANA DB configuration."""
 
 import os
+from enum import Enum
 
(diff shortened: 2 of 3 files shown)


reana_db/models.py
```diff
@@ -711,8 +712,12 @@ def _update_cpu_quota(workflow):
         RunStatus.failed,
         RunStatus.stopped,
     ]:
-        _update_cpu_quota(workflow)
-        _update_disk_quota(workflow)
+        quota_update_policy = os.getenv("REANA_WORKFLOW_TERMINATION_QUOTA_UPDATE_POLICY", "cpu,hdd").split(",")
```
tiborsimko, line 715: As we call it `disk`, please change `hdd` to `disk` everywhere

tiborsimko: Also, it would be good to move `os.getenv()` to `reana_db/config.py` next to the others.

tiborsimko: BTW the check might be needed in more places. For my tests, I ended up using the following to turn off all cpu and disk quota updating:

```diff
diff --git a/reana_db/models.py b/reana_db/models.py
index c9c007e..87e54ad 100644
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -583,6 +583,7 @@ class Workflow(Base, Timestamp, QuotaBase):
 
     def get_workspace_disk_usage(self, summarize=False, search=None):
         """Retrieve disk usage information of a workspace."""
+        return [{"name": "", "size": 0}]
         from functools import partial
 
         if not summarize:
@@ -684,10 +685,12 @@ def workflow_status_change_listener(workflow, new_status, old_status, initiator)
     from .database import Session
 
     def _update_disk_quota(workflow):
+        return
         update_users_disk_quota(user=workflow.owner)
         store_workflow_disk_quota(workflow)
 
     def _update_cpu_quota(workflow):
+        return
         terminated_at = workflow.run_finished_at or workflow.run_stopped_at
         if workflow.run_started_at and terminated_at:
             cpu_time = terminated_at - workflow.run_started_at
diff --git a/reana_db/utils.py b/reana_db/utils.py
index e9ccec7..e96f29c 100644
--- a/reana_db/utils.py
+++ b/reana_db/utils.py
@@ -220,6 +220,7 @@ def update_users_disk_quota(user=None, bytes_to_sum=None):
     :type bytes_to_sum: int
 
     """
+    return
     from reana_commons.utils import get_disk_usage
 
     from reana_db.config import DEFAULT_QUOTA_RESOURCES
@@ -249,6 +250,7 @@ def update_users_disk_quota(user=None, bytes_to_sum=None):
 
 def get_disk_usage_or_zero(workspace_path):
     """Get disk usage for the workspace if exists, zero if not."""
+    return 0
     from reana_commons.utils import get_disk_usage
     from reana_commons.errors import REANAMissingWorkspaceError
 

```

PR #152, opened 2021-10-27
config: adapt quota update policy to helm values

mvidalgarcia:
Adapt assignment logic to avoid a list with an empty string `[""]`
when the policy coming from Helm is empty.

closes reanahub/reana#562


audrium (approved):
LGTM! works nicely!

PR #161, opened 2022-02-14
models: allow workflow to fail from queued status

VMois:
closes reanahub/reana-server#436

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 Changes
 =======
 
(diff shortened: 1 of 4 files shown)


VMois:
*note(non-blocking):* this is an example of breaking boundaries between codebases. This is not convenient to store workflow transition statuses in `models.py` in `r-db`. A better place, in my opinion, is `r-workflow-controller` so all changes related to workflow schedules are contained in one PR.

audrium (approved):
Looks good!

PR #165, opened 2022-03-07
models: add `launcher_url` column to `Workflow` table

mvidalgarcia:
closes reanahub/reana-ui#222


audrium (approved):
LGTM! Alembic works as expected
What is missing is a note inside `CHANGES.rst`

Issue #172, opened 2022-06-17
models: new DB table to hold workspace file retention rules


tiborsimko:
Design a new DB table for keeping retention rules following RFC https://github.com/reanahub/reana-client/issues/610

Example:

| workflow_id (FK) | workspace_files     | retention_days  | apply_on            |
| ---------------- | ------------------- | --------------- | ------------------- |
| 1111-2232-3333   | myfetched/**/*.root | 1               | NULL                |
| 1111-2232-3333   | myfiltered/*.root   | 7               | NULL                |
| 1111-2232-3333   | myfits/             | 30              | NULL                |
| 1111-2232-3333   | myplots/*.eps       | 365             | NULL                |
| 1111-2232-3333   | myplots/*.png       | 365             | 2023-05-02 00:00:01 |

This table will be populated by parsing `workspace.retention_days` from `reana.yaml` when the workflow is created.

(Side-comment: and, in case a user reruns a workflow, they may modify `reana.yaml` as well, so please beware that these entries can change after DB rows are created. IOW we need both nice CREATE and nice UPDATE methods to manipulate this table for a given workflow.)

This table will then be used by the nightly workspace retention rule applier daemon that will read the rows, and, depending on the workflow status, process some of them (and delete the files), and skip some of them (until their time will come).

Note that the `apply_on` date time column is illustrating one way how we could avoid processing too many rows unnecessarily every night. The defalut value is NULL, meaning the row was not processed yet. Example: the workflow run 1111-2222-3333 finishes on May 1st. The daemon will first process this workflow run on May 2nd. It will go through each rule and set `apply_on` date time stamp appropriately, e.g. for `myplots/*.png` it will be set to `2023-05-02 00:00:01`. And when the daemon will run the next day on May 3rd, it won't need to touch only those rows `WHERE NOW() > apply_on`, so to speak.

The details on the daemon's work will be part of another ticket, but it is good to think about these usage scenarios whilst designing the DB model table in order to avoid doing many DB model changes later.

The expected outcome if this issue therefore is:

- think about a DB workspace file retention rule storage table and either confirm the above high-level design or propose an alternative;
- create a new DB model table (with efficient indexes since it may grow a lot);
- define nice methods to insert (by workflow submitter) and update (by both workflow update by user or by the future retention rule processing daemon);
- create new DB alembic upgrade recipe.

VMois:
In addition, it would be good to populate our REANA dev wiki with a page on how to work with Alembic.

[...]

PR #174, opened 2022-06-22
models: add workspace retention rules table

audrium:
closes https://github.com/reanahub/reana-db/issues/172

To test:

- `kubectl exec -i -t deployment/reana-server -c rest-api -- invenio shell`:
```
from reana_db.models import Workflow, WorkspaceRetentionRule, WorkspaceRetentionRuleStatus
from reana_db.utils import update_workspace_retention_rules
from reana_db.database import Session
from datetime import datetime
s = Session()

workflow_id = Workflow.query.first().id_
workspace_files = "myfetched/**/*.root"
retention_days = 1
apply_on = None

# Insert a single rule
wr = WorkspaceRetentionRule(workflow_id=workflow_id, workspace_files=workspace_files, retention_days=retention_days)
s.add(wr)
s.commit()

# Update a single rule
wr.status = WorkspaceRetentionRuleStatus.inactive
s.add(wr)
s.commit()

# Inspect audit logs
wr.audit_logs.all()

# Insert multiple rules while creating the workflow
w = Workflow.query.first()
retention_rules = [
    {
        "workspace_files": "myfiltered/*.root",
        "retention_days": 7
    },
    {
        "workspace_files": "myfits/",
        "retention_days": 30
    },
]
w.set_workspace_retention_rules(retention_rules)

# Update multiple rules
wr_all = WorkspaceRetentionRule.query.filter(WorkspaceRetentionRule.workflow_id == workflow_id).all()
for rule in wr_all:
    rule.status = WorkspaceRetentionRuleStatus.applied
    s.add(rule)
s.commit()

# Update all workflow rules
workflow = Workflow.query.first()
workflow.inactivate_workspace_retention_rules()
workflow.activate_workspace_retention_rules()

# Try to transit to a wrong status
update_workspace_retention_rules(workflow, WorkspaceRetentionRuleStatus.created)
```

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -16,11 +16,13 @@ import logging
 import uuid
 from datetime import datetime
 from functools import reduce
+from typing import Dict, List
 
 from reana_commons.config import (
     MQ_MAX_PRIORITY,
     REANA_MAX_CONCURRENT_BATCH_WORKFLOWS,
     REANA_RUNTIME_KUBERNETES_KEEP_ALIVE_JOBS_WITH_STATUSES,
+    WORKFLOW_TIME_FORMAT,
 )
 from reana_commons.errors import REANAValidationError
 from reana_commons.utils import get_disk_usage
@@ -32,6 +34,7 @@ from sqlalchemy import (
     Enum,
     Float,
     ForeignKey,
+    Integer,
     String,
     Text,
     UniqueConstraint,
@@ -59,6 +62,7 @@ from reana_db.utils import (
     store_workflow_disk_quota,
     update_users_disk_quota,
     update_workflow_cpu_quota,
+    update_workspace_retention_rules,
 )
 
 
@@ -469,13 +473,15 @@ class Workflow(Base, Timestamp, QuotaBase):
     git_provider = Column(String(255))
     launcher_url = Column(String)
     owner = relationship("User", backref="workflow")
-
     sessions = relationship(
         "InteractiveSession",
         secondary="__reana.workflow_session",
         lazy="dynamic",
         backref="workflow",
         cascade="all, delete",
+    )
+    retention_rules = relationship(
+        "WorkspaceRetentionRule", backref="workflow", lazy="dynamic"
     )
 
     __table_args__ = (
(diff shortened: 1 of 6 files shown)


reana_db/models.py
```diff
@@ -812,6 +833,73 @@ def __repr__(self):
...
+    action = Column(
+        Enum(WorkspaceRetentionRuleStatus), primary_key=True, nullable=False
+    )
+
+    def __repr__(self):
+        """Workspace Retention audit log string representation."""
+        return f"<WorkspaceRetentionAuditLog {self.workspace_retention_rule_id} {self.action}>"
+
+
+class WorkspaceRetention(Base):
```
tiborsimko, line 867: We have three new classes with this PR:

- class WorkspaceRetentionRuleStatus(enum.Enum):
- class WorkspaceRetentionAuditLog(Base):
- class WorkspaceRetention(Base):

Quick impressions seeing these names: the first is nicely self-understandable, the third seem to be unfinished (retention what?) as if calling for a final noun (WorkspaceRetentionRule?), and the second is perhaps too businessy if we include "audit" (WorkspaceRetentionLogs?) and may perhaps clash with the AuditLog table that we have already for some user actions. 

(BTW it may be a good idea to reuse the existing AuditLog one, so that we have centralised all the "auditable actions" in one place... but that would need more changes.)

reana_db/models.py
```diff
@@ -627,6 +633,42 @@ def get_workspace_disk_usage(self, summarize=False, search=None):
 
         return [{"name": "", "size": disk_usage}]
 
+    def set_workspace_retention_rules(self, rules: List[Dict[str, str]]):
+        """Set workspace retention rules for the workflow."""
+        from .database import Session
+
+        for rule in rules:
```
mdonadoni, line 640: What do you think about having a limit on the number of rules? (in any case, we can do this in another PR)

reana_db/utils.py
```diff
@@ -351,6 +352,35 @@ def update_users_cpu_quota(user=None) -> None:
...
+            continue
+        if not rule.can_transition_to(status):
+            raise Exception(
+                f"Cannot transition workspace retention rule {rule.id_} "
+                f"from status {rule.status} to {status}."
+            )
+        if status == WorkspaceRetentionRuleStatus.active:
+            rule.apply_on = datetime.today().replace(
+                hour=23, minute=59, second=59
+            ) + timedelta(days=rule.retention_days)
```
mdonadoni, line 378: Should we also set `apply_on` to `None` when inactivating rules?

mdonadoni (commented):
The PR works well, good job! I have left just a few comments

mdonadoni:
Late comment: I think it would be a good idea to add a few tests to this PR!

reana_db/models.py
audrium: I think it could be tackled as part of [validation issue](https://github.com/reanahub/reana-server/issues/507). I've linked this comment there in the description

reana_db/models.py
```diff
@@ -812,6 +856,106 @@ def __repr__(self):
...
+        "WorkspaceRetentionAuditLog", backref="workspace_retention_rule", lazy="dynamic"
+    )
+
+    def can_transition_to(self, next_status):
+        """Whether the provided retention rule can transition to the next status."""
+        current_transition = (self.status, next_status)
+        return current_transition in ALLOWED_WORKSPACE_RETENTION_RULE_STATUS_TRANSITIONS
+
+    @property
+    def serialize(self):
```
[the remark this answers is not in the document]
mdonadoni: > Do you have something specific in mind? Probably better to create an issue about this

I was not thinking about anything in particular, we can create an issue to investigate how to proceed!

mdonadoni (approved):
Works well, great job! :rocket:

Issue #173, opened 2022-06-20
models: incorrect run number assignment after many restarts


mdonadoni:
How to reproduce:

1. Execute one workflow (e.g. a demo example)
2. Restart the workflow ten times
3. The workflow runs will have the following run numbers:
    - `workflow.1` (initial run)
    - `workflow.1.1` (first restart)
    - `workflow.1.2` (second restart)
    - ...
    - `workflow.1.9` (ninth restart)
    - `workflow.2` (tenth restart)

However, the tenth restart should be `workflow.1.10`.

At this point, it is also not possible to create a new workflow (`reana-client create ...`) with the same name. The workflow creation will fail given that `workflow.2` is already being used:
```
(psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "_user_workflow_run_uc"
DETAIL:  Key (name, owner_id, run_number)=(workflow, 00000000-0000-0000-0000-000000000000, 2) already exists.
```

The cause of the issue is that run numbers are stored as floats and the new run numbers are generated in this way:
https://github.com/reanahub/reana-db/blob/dd1b60ede1dbfe9f7f98c60918fb5c3bac4bade5/reana_db/models.py#L566

One possible solution would be to store each run number as two integers and update the rest of the codebase accordingly.

audrium:
Temporarily closed with https://github.com/reanahub/reana-db/pull/176. Reopening for the future to implement a different way how to store the `run_number` value in the DB

audrium:
Created a new issue to solve the remaining limit of nine restarts https://github.com/reanahub/reana-db/issues/186

PR #176, opened 2022-06-27
models: limit the number of restarts of a workflow

mdonadoni:
Considering issue #173, the limit is currently set to permit at most
nine restarts.

How to test: see #173


audrium (approved):
Works nicely!

PR #178, opened 2022-07-19
models: always calculate disk usage for Workflow.get_workspace_disk_usage

VMois:
closes reanahub/reana-server#521

## How to test?

0. Try reproducing an error from the issue with disabled quota.

```diff
diff --git a/helm/reana/values.yaml b/helm/reana/values.yaml
index 487683b..e2b51ed 100644
--- a/helm/reana/values.yaml
+++ b/helm/reana/values.yaml
@@ -142,7 +142,7 @@ traefik:

 # Quota
 quota:
-  enabled: true
+  enabled: false
   periodic_update_policy: "{{ .Values.quota.disk_update }}"
   workflow_termination_update_policy: "{{ .Values.quota.termination_update_policy }}"
   # backward compatibility
```

1. Keep disabled quotas. Checkout the PR, run `reana-dev git-submodule --update`
2. Try to reproduce the issue again, it should not be possible

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ Version 0.9.0 (UNRELEASED)
 - Adds new ``WorkspaceRetentionRule`` table to store workspace file retention rules.
 - Adds a limit to the number of times a workflow can be restarted.
 - Changes percentage ranges for calculating the quota health status.
+- Changes ``Workflow.get_workspace_disk_usage`` to calculate disk usage and not use the usage values from database.
 
 Version 0.8.2 (2022-02-23)
 ---------------------------
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -609,29 +609,14 @@ class Workflow(Base, Timestamp, QuotaBase):
         """Retrieve disk usage information of a workspace."""
         from functools import partial
 
-        if not summarize:
-            # size break down per directory so we can't query DB (`r-client du`)
-            return get_disk_usage(
-                self.workspace_path,
-                summarize,
-                search,
-                to_human_readable_units=partial(
-                    ResourceUnit.human_readable_unit, ResourceUnit.bytes_
-                ),
-            )
-
-        disk_usage = self.get_quota_usage().get("disk", {}).get("usage", {})
-        if not disk_usage:
-            # recalculate disk workflow resource
-            workflow_resource = store_workflow_disk_quota(self)
-            disk_usage = dict(
-                raw=workflow_resource.quota_used,
-                to_human_readable_units=ResourceUnit.human_readable_unit(
-                    ResourceUnit.bytes_, workflow_resource.quota_used
-                ),
-            )
-
-        return [{"name": "", "size": disk_usage}]
+        return get_disk_usage(
+            self.workspace_path,
+            summarize,
+            search,
+            to_human_readable_units=partial(
+                ResourceUnit.human_readable_unit, ResourceUnit.bytes_
+            ),
+        )
 
     def set_workspace_retention_rules(self, rules: List[Dict[str, str]]):
         """Set workspace retention rules for the workflow."""


mdonadoni (commented):
The PR works well!

However, after these changes the disk quota is never used to avoid calculating the disk usage. Also, if the disk quota usage is not present, it is not updated anymore when a user calls `reana-client du`. Is this acceptable? cc @tiborsimko @audrium

CHANGES.rst
```diff
@@ -8,6 +8,7 @@ Version 0.9.0 (UNRELEASED)
 - Adds new ``WorkspaceRetentionRule`` table to store workspace file retention rules.
 - Adds a limit to the number of times a workflow can be restarted.
 - Changes percentage ranges for calculating the quota health status.
+- Changes ``Workflow.get_workspace_disk_usage`` to calculate disk usage and not use the usage values from database.
```
tiborsimko, line 11: Hmm, I'm not sure whether it is OK to _always_ calculate the workspace disk usage without caching result. This could lead to slow REST API replies to commands like `du -w myoldworkflow` even if those old files didn't change much in years...

Besides, how would this influence commands like `list --include-workspace-size`? Would we do everything live as well? (Here the REST API performance is much more important than for single `du` command.)

I'd say `reana-client du` or system-wide `reana-resource-quota-update` cronjob or `reana-client list --include-workspace-size` are all kinds of commands that could trigger disk usage getter, and then if the result is cached in the database, all those commands could return detected values.

E.g. currently I'm getting `-1` in the size column locally on a fresh installation, even if I ask for one disk usage explicitly before:

```console
$ reana-client du -w root6-roofit-snakemake-kubernetes -s
SIZE      NAME
1556480   .
$ reana-client list --include-workspace-size
NAME                                RUN_NUMBER   CREATED               STARTED               ENDED                 STATUS     SIZE
root6-roofit-snakemake-kubernetes   1            2022-08-08T11:45:57   2022-08-08T11:46:10   2022-08-08T11:46:41   finished   -1
root6-roofit-serial-kubernetes      1            2022-08-08T11:45:24   2022-08-08T11:45:36   2022-08-08T11:45:51   finished   -1
root6-roofit-cwl-kubernetes         1            2022-08-08T11:44:49   2022-08-08T11:45:03   2022-08-08T11:45:18   finished   -1
```

I guess we may want to polish the du/list/cronjob chain and reuse of results more widely?

VMois: > Besides, how would this influence commands like list --include-workspace-size? Would we do everything live as well? (Here the REST API performance is much more important than for single du command.)

The performance will not be affected. `du` command in this PR doesn't share a code. The `list --include-workspace-size` command uses quotas, and if they do not exist, it returns `-1`. If quotas are disabled and we want to cache results, `du` and `list --include-workspace-size` commands will be outdated. 

> I guess we may want to polish the du/list/cronjob chain and reuse of results more widely?

Yes, we should. Our quota codebase is messy and hard to modify and understand. Should I look into ways to improve it? I can prepare the documentation for the REANA dev wiki describing how current quotas work. From there we can decide what to do.

tiborsimko: OK, I confirm that `list` is OK. Hence we can merge this fix. And we can create a separate issue for the `-1` bit.  (That we should at the very least document in `reana-client list --help`, and at the very most update so that people can consult disk space even if quotas are disabled.)

tiborsimko:
> Also, if the disk quota usage is not present, it is not updated anymore when a user calls `reana-client du`. Is this acceptable? cc @tiborsimko @audrium

I guess it is OK, provided that nightly cronjob updates the disk quota usage numbers once per day.

In this way, `du` means "live lookup", and `list` means "DB quota usage lookup", so to speak. Which seems nicely consistent. And we'll create a separate issue for better caching du/list/cronjob disk usage results in the DB and share amongst them (fixing among others the `-1` behaviour when quotas are disabled, as discussed elsewhere)

tiborsimko (approved):
Works well, and related issues (such as dealing with `list` output being `-1` in non-quota-conditions) are to be ticketised for later.

PR #183, opened 2022-10-06
models: add `pending` status for retention rules

mdonadoni:
Closes reanahub/reana-server#526


audrium (approved):
Works great! We could also drop a new line in `CHANGES.rst`

PR #184, opened 2022-11-14
utils: check user is the owner of the workflow

mdonadoni:
Check that the current user is the owner of the workflow when searching by UUID.


audrium (approved):
Works nicely

Issue #193, opened 2023-01-16
quotas: improve performance of quota updater


mdonadoni:
The periodic quota updater takes a long time to complete when dealing with many workflows and/or with big workspaces.

These are some possible improvements:
1. Currently, `Session.commit()` is called once for every workflow in `store_workflow_disk_quota` ([here](https://github.com/reanahub/reana-db/blob/7998fd24821459ff91fe0f45a560cc759254ed70/reana_db/utils.py#L462) and [here](https://github.com/reanahub/reana-db/blob/7998fd24821459ff91fe0f45a560cc759254ed70/reana_db/utils.py#L470)). Instead, it should be called only once after all the quota updates have been calculated. This same improvements can also be applied to the other utility functions that are used to update the disk/cpu quotas of workflows and users. In particular, I have noticed that `Session.commit()` becomes slower the more workflows are loaded from the database.
2. The disk quota usage of all the workflows are recalculated, even though many workflows have not changed since the last execution of the quota updater. One possible solution to this would be to consider only workflows that are "dirty", that is workflows whose workspace might have changed since the last quota update (e.g. a session has been opened, a file has been deleted/uploaded, etc.)

mdonadoni:
See also https://github.com/reanahub/reana-db/pull/200#issuecomment-1693415203

PR #200, opened 2023-08-25
quota: improve performance of quota updater

mdonadoni:
quota: improve performance of quota updater

Improve the performance of the quota updater by "expunging" all
workflows before starting to update quotas, as the Workflow table does
not need to be modified, thus making commits much faster. Also refactor
the various update functions to make their behaviour consistent with
each other.

Partially addresses https://github.com/reanahub/reana-db/issues/193

### DIFF
--- a/reana_db/utils.py
+++ b/reana_db/utils.py
@@ -206,6 +206,65 @@ def _get_workflow_by_uuid(workflow_uuid, user_uuid):
     return workflow
 
 
+class Timer:
+    """Timer to time events and log periodic progress."""
+
+    def __init__(self, name=None, total=None, periodic_delta=100) -> None:
+        """Initialise new Timer."""
+        self.name = name
+        self.total = total
+        self.periodic_delta = periodic_delta
+        self.count = 0
+        self.start = datetime.now()
+
+    def elapsed(self) -> float:
+        """Elapsed time since the creation of the Timer, in seconds."""
+        diff = datetime.now() - self.start
+        return diff.total_seconds()
+
+    def estimated_total(self) -> float:
+        """Estimated total time, in seconds."""
+        if not self.total or not self.count:
+            return 0
+        return self.elapsed() * self.total / self.count
+
+    def per_event(self) -> float:
+        """Time per event, in seconds."""
+        if self.count == 0:
+            return 0
+        return self.elapsed() / self.count
+
+    def log_progress(self) -> None:
+        """Log progress of events."""
+        progress = ""
+        if self.name:
+            progress = f"{self.name} "
+        progress += f"progress: {self.count}"
+        if self.total:
+            progress += f"/{self.total}"
+        progress += (
+            f"  elapsed: {self.elapsed():.3f}s"
+            f"  est.total: {self.estimated_total():.3f}s"
+            f"  per event: {self.per_event():.3f}s"
+        )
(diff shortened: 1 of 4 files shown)


mdonadoni:
Small note for this PR: this PR maintains the current approach of having one commit per workflow/user, but this will not scale with an even bigger number of workflows and/or users.
Ideas for improvements:
- use bulk operations
- consider only workflows that were actually "modified" and not all of them
- run the quota updates asynchronously

tiborsimko:
Some thoughts on the possible _future_ options:

> use bulk operations

If we make only one commit at the end of the session, and we would have to update many rows, and this last commit would fail for a reason-or-another, how would the rollback situation look? Might not be so interesting as the other two approaches?

> consider only workflows that were actually "modified" and not all of them

This would be definitely good to do longer term, i.e. process only "modified" workspaces, and having an option to process "everything" once in a blue moon to catch up any possible troubles.

However for operations such as `rm` we wouldn't know just by checking the workspace contents whether some file was removed from notebook session terminal. So we'd have to take into account not only the workflow file manipulations, but also when a terminal might have been opened/closed on some workspace.

> run the quota updates asynchronously

Yes, it might be good to have another queue where we register "please process quota for this workflow" kind of requests and have another daemon consumer process running permanently in the background that would consume these messages and doing the necessary quota updates as they arrive.  In this way we would also address the previous item about doing only those updates that are actually needed. (Even though there may still be corner cases such a a user leaving a notebook terminal opened for many days without closing.)  This requires quite some development though, but we might want to ticketise the ideas for the future!

reana_db/utils.py
```diff
@@ -477,3 +559,23 @@ def store_workflow_disk_quota(
...
+    # logs and reana_specification are not loaded to avoid consuming
+    # huge amounts of memory
+    workflows = Workflow.query.options(
+        defer(Workflow.logs), defer(Workflow.reana_specification)
+    ).all()
+    # We expunge all the workflows, as they will not be modified when updating the quotas.
+    # This makes `Session.commit()` much faster
+    for workflow in workflows:
+        Session.expunge(workflow)
+    timer = Timer("Workflow disk", total=len(workflows))
```
tiborsimko, line 578: Cosmetics: the log messages currently look like:

```
2023-08-25 14:22:22,682 | root | MainThread | INFO | Workflow disk progress: 8/8  elapsed: 0.066s  est.total: 0.066s  per event: 0.008s
2023-08-25 14:22:22,703 | root | MainThread | INFO | User disk progress: 3/3  elapsed: 0.019s  est.total: 0.019s  per event: 0.006s
Users disk quota usage updated successfully.
2023-08-25 14:22:22,721 | root | MainThread | INFO | Workflow CPU progress: 8/8  elapsed: 0.015s  est.total: 0.015s  per event: 0.002s
2023-08-25 14:22:22,738 | root | MainThread | INFO | User CPU progress: 3/3  elapsed: 0.016s  est.total: 0.016s  per event: 0.005s
Users cpu quota usage updated successfully.
```

We may want to improve the INFO message (it's not "CPU progress"!) by using something like: 

- "Workflow disk quota usage update progress: 8/8"

and clarify the final message:

- "Disk quota usage updated successfully for all users and workflows."

tiborsimko (approved):
Works well locally :+1:

PR #201, opened 2023-09-01
quota: avoid accessing disk when updating user disk quotas

mdonadoni:
Partially addresses #193

### DIFF
--- a/reana_db/utils.py
+++ b/reana_db/utils.py
@@ -299,6 +299,10 @@ def update_users_disk_quota(
 ) -> None:
     """Update users disk quota usage.
 
+    User disk quota usage will be calculated from the individual workflow disk quota
+    usage numbers, so this function should be typically called only after
+    ``update_workflows_disk_quota()``.
+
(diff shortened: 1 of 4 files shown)


reana_db/utils.py
```diff
@@ -299,6 +299,9 @@ def update_users_disk_quota(
 ) -> None:
     """Update users disk quota usage.
 
+    User disk quotas will be calculated from workflow disk quotas,
+    so the latter should be updated before the former.
```
tiborsimko, line 303: Cosmetics: perhaps add the explicit function name for clarity, e.g.  "User disk quota usage will be calculated from the individual workflow disk quota usage numbers, so this function should be typically called only after `update_workflows_disk_quota()`."

tiborsimko (approved):
Works nicely :+1: Left one cosmetic comment

reana_db/utils.py
tiborsimko: Please also write an entry into `CHANGES.rst` about performance improvements for disk quota calculations. (The last PR also omitted news item about this.)

PR #207, opened 2023-10-26
models: add workflow run sharing table

### LINKED ISSUE #206: model: design workflow run sharing tables
For the workflow sharing sprint https://github.com/reanahub/reana/issues/751 we shall need DB model changes to express that Alice shared her workflow run myanalysis 42 with user Bob.

The DB design should be flexible enough to hint at various "user roles" e.g. Bob is only a R/O reader, but we should pave the way towards possible future more roles (e.g. someone who can restart workflows, e.g. someone who could modify some parts in R/W manner in Alice's name, etc.)

The goal of this issue it to design the new DB tables and prepare an Alembic recipe for upgrades of existing deployments.

DaanRosendal:
Adds two tables to the database. The `user_workflow` table makes it possible to verify whether a specific workflow has been shared with a user. The `role` table stores various roles, such as 'read-only' and 'read-write.'

Closes #206

---

This PR can be tested using the usual DB update testing procedures described here: https://github.com/reanahub/reana-db/pull/128#issue-878947607

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -11,8 +11,8 @@
 from __future__ import absolute_import
 
 import enum
-import math
 import logging
+import math
 import uuid
 from datetime import datetime
 from functools import reduce
@@ -26,6 +26,21 @@ from reana_commons.config import (
 )
 from reana_commons.errors import REANAValidationError
 from reana_commons.utils import get_disk_usage
+from reana_db.config import (
+    DB_SECRET_KEY,
+    DEFAULT_QUOTA_LIMITS,
+    DEFAULT_QUOTA_RESOURCES,
+    LIMIT_RESTARTS,
+    WORKFLOW_TERMINATION_QUOTA_UPDATE_POLICY,
+)
+from reana_db.utils import (
+    build_workspace_path,
+    store_workflow_disk_quota,
+    update_users_cpu_quota,
+    update_users_disk_quota,
+    update_workflow_cpu_quota,
+    update_workspace_retention_rules,
+)
 from sqlalchemy import (
     BigInteger,
     Boolean,
@@ -43,12 +58,14 @@ from sqlalchemy import (
     func,
     or_,
 )
+from sqlalchemy.dialects.postgresql import ARRAY
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import relationship
 from sqlalchemy_utils import EncryptedType, JSONType, UUIDType
 from sqlalchemy_utils.models import Timestamp
 from sqlalchemy_utils.types.encrypted.encrypted_type import AesEngine
+<<<<<<< HEAD
 from sqlalchemy.dialects.postgresql import ARRAY
 
 from reana_db.config import (
@@ -67,6 +84,8 @@ from reana_db.utils import (
     update_workspace_retention_rules,
 )
 
+=======
+>>>>>>> 26a0532 (models: add workflow run sharing tables)
 
 convention = {
     "ix": "ix_%(column_0_label)s",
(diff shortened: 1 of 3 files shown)


reana_db/models.py
```diff
@@ -147,6 +145,12 @@ class User(Base, Timestamp, QuotaBase):
     username = Column(String(length=255))
     tokens = relationship("UserToken", backref="user_", lazy="dynamic")
     workflows = relationship("Workflow", backref="user_", lazy="dynamic")
+    shared_workflows = relationship(
+        "Workflow",
+        secondary="__reana.user_workflow",
+        backref="shared_users",
+        lazy="dynamic",
+    )
```
mdonadoni, line 153: While looking into many-to-many relationships in SQLAlchemy, I have stumbled upon the warning in this section:  https://docs.sqlalchemy.org/en/13/orm/basic_relationships.html#association-object

It might be a good idea to make this relationship `shared_workflows` read-only (but I do not know whether this affects other parts of the workflow sharing code, so let me know if this is not feasible).

reana_db/models.py
```diff
@@ -1151,3 +1155,28 @@ class QuotaHealth(enum.Enum):
...
+    __tablename__ = "user_workflow"
+    __table_args__ = {"schema": "__reana"}
+
+    workflow_id = Column(UUIDType, ForeignKey("__reana.workflow.id_"), primary_key=True)
+    user_id = Column(UUIDType, ForeignKey("__reana.user_.id_"), primary_key=True)
+    access_type = Column(Enum(AccessType), default=AccessType.read, nullable=False)
+    message = Column(Text)
+    valid_until = Column(DateTime)
+
+    user = relationship("User", backref="user_workflows")
```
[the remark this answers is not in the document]
mdonadoni: (PS: also the UserWorkflow table name could be changed to something less generic, up to you!)

reana_db/models.py
```diff
@@ -1151,3 +1155,28 @@ class QuotaHealth(enum.Enum):
...
+class UserWorkflow(Base):
+    """UserWorkflow table."""
+
+    __tablename__ = "user_workflow"
+    __table_args__ = {"schema": "__reana"}
+
+    workflow_id = Column(UUIDType, ForeignKey("__reana.workflow.id_"), primary_key=True)
+    user_id = Column(UUIDType, ForeignKey("__reana.user_.id_"), primary_key=True)
+    access_type = Column(Enum(AccessType), default=AccessType.read, nullable=False)
+    message = Column(Text)
```
mdonadoni, line 1175: user_email_to_share_with

mdonadoni, line 1175: minor: we could consider limiting the length of this field also in the database schema by using `String(a1234...)`, up to you!

reana_db/models.py
mdonadoni: Also, we could maybe change the name to `workflows_shared_with_me` or something similar? What do you think?
The current name is ambiguous, as it's not clear whether these are the workflows shared _with_ me or _by_ me

reana_db/models.py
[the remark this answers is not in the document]
DaanRosendal: With my recent changes, the User table now has:

1. `user.workflows`
2. `user.workflows_shared_with_me`

As for 3, I decided to remove it since it is not being used. I did not change the `UserWorkflow` table name because I think the current name is fine, it follows the standard many to many relationship naming convention

The changes were applied to the following superseding PR: https://github.com/reanahub/reana-db/pull/216

reana_db/models.py
mdonadoni: I think that `viewonly=True` is not needed anymore now that you have removed `user_workflows` , but let's keep it like this for now

mdonadoni:
Included in https://github.com/reanahub/reana-db/pull/216

PR #212, opened 2023-11-21
utils: add `include_shared_workflows` parameter

DaanRosendal:
Adds new `include_shared_workflows` parameter to optionally retrieve
unowned workflows that are shared with the user.

Closes reanahub/reana-client#687

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -5,6 +5,7 @@ The list of contributors in alphabetical order:
 
 - `Audrius Mecionis <https://orcid.org/0000-0002-3759-1663>`_
 - `Camila Diaz <https://orcid.org/0000-0001-5543-797X>`_
+- `Daan Rosendal <https://orcid.org/0000-0002-3447-9000>`_
 - `Diego Rodriguez <https://orcid.org/0000-0003-0649-2002>`_
 - `Dinos Kousidis <https://orcid.org/0000-0002-4914-4289>`_
 - `Jan Okraska <https://orcid.org/0000-0002-1416-3244>`_
--- a/reana_db/alembic/versions/20231102_2003_4543cde7c9fb_workflow_sharing.py
+++ b/reana_db/alembic/versions/20231102_2003_4543cde7c9fb_workflow_sharing.py
@@ -1,0 +1,46 @@
+"""Workflow sharing.
+
+Revision ID: 4543cde7c9fb
+Revises: 377cfbfccf75
+Create Date: 2023-11-02 20:03:01.234086
+
+"""
+import sqlalchemy as sa
+import sqlalchemy_utils
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "4543cde7c9fb"
+down_revision = "377cfbfccf75"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    """Upgrade to 4543cde7c9fb revision."""
+    op.create_table(
+        "user_workflow",
+        sa.Column(
+            "workflow_id", sqlalchemy_utils.types.uuid.UUIDType(), nullable=False
+        ),
+        sa.Column("user_id", sqlalchemy_utils.types.uuid.UUIDType(), nullable=False),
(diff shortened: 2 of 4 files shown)


mdonadoni (commented):
I have not checked whether the PR works, but I just wanted to leave a general comment about it as I think these changes can be somewhat problematic.

Regarding searching a workflow by UUID, I think it's fine to have the `include_shared_workflows` flag, as the workflow is uniquely identified in any case. This means that:
- UUID of a shared workflow but `include_shared_workflows` is false => workflow is not returned
- UUID of a shared workflow and `include_shared_workflows` is true => workflow is returned

This is quite clear and is a very basic version of RBAC.

However, searching by name is somewhat broken. In particular, if `include_shared_workflows` is set to true, then workflows owned by the user and workflows shared with the user are all mixed together. This means there can now be multiple workflows with the same name _and same run number_, owned by different users!

First of all, this breaks parts of the code, for example:
```python
workflow = (
    Workflow.query.outerjoin(
        UserWorkflow, UserWorkflow.workflow_id == Workflow.id_
    )
    .filter(
        (Workflow.name == workflow_name)
        & (Workflow.run_number == run_number)
        & (
            (Workflow.owner_id == user_uuid)
            | (UserWorkflow.user_id == user_uuid)
        )
    )
    .one_or_none()
)
```
This query can now potentially return multiple workflows, so `one_or_none` will fail.

Furthermore, this change has also a very confusing semantic. Up until now:
- `my-roofit-analysis` => Workflow owned by me with that name and biggest run number (or error if it does not exist)
- `my-roofit-analysis.1` => Workflow owned by me with that same name and major run number, but biggest minor run number (or error if it does not exist)
- `my-roofit-analysis.1.12` => Workflow owned by me is uniquely identified and it is returned (or error if it does not exist)

After this change, any of those will now either refer to an owned workflow or a shared workflow, without the user being able to know which workflow he is referring to. The user is also not able to specify which workflow he **wants** to refer to, as there is no way for him to specify a "fully-qualified" workflow name.

I see a few options, but feel free to suggest more:
1. Do not allow users to refer to shared workflows by name, but only by ID. When searching by name, only consider owned workflows.
2. Think about a way to let the users specify a "fully-qualified" name that uniquely identifies workflows. As an example, something like `other-user/workflow-shared-with-me.1.2`.

Option 1 should be easy to implement, but it's quite short-sighted as it makes the user experience worse, given that they will need to use UUID to get the details of shared workflows. This will be even worse when we will add read-write sharing options in the future.

Option 2 is better regarding user experience, but I think it's a huge undertaking, given that we would probably need to change lots of REANA code. Not only that, we are currently not using usernames in REANA and we are not letting people choose and change them, so we cannot use them to identify users. The alternative would be to use the email, so a fully qualified name would be similar to `a@a.com/workflow-shared-with-me.1.2`. In any case, this needs more thought and discussion.

What do you think?

What does the team think? @tiborsimko @giuseppe-steduto

giuseppe-steduto:
These are very valid concerns that indeed require a better discussion.

Option 1: I was trying to put myself in the shoes of other users, and while it may make commands appear less intuitive, it essentially boils down to obtaining the UUID from the `list` command. Although this might seem like a minor inconvenience, I think this could be quickly become moderately annoying to users if the workflow has to be interacted with a lot.

Option 2: I haven't really explored the possibility, but maybe a way to make fewer changes to the code is to enable a fully-qualified name made like `email/my-analysis.1.2`, but default to the user's workflow when only the workflow name is given. In this way,
`my-roofit-analysis`, `my-roofit-analysis.1`, and `my-roofit-analysis.1.12` would still work as they are right now, but `user@example.com/my-roofit-analysis.1` for example would refer to the workflow *owned by `user@example.com`* with that same name and major run number, but biggest minor run number (or error if it does not exist). In this way, there wouldn't be ambiguity.

Further discussion within the team is in any case needed to weigh the pros and cons of these options well.

mdonadoni:
Included in https://github.com/reanahub/reana-db/pull/216, the discussion can continue there

PR #213, opened 2023-11-23
models: improve indexes

### LINKED ISSUE #211: global: monitor database queries
We should investigate what options exist to monitor and profile database queries.

This will allow us to find slow queries (if any) which we can optimise (e.g. creating new database indexes, rewriting a different query, etc.)

mdonadoni:
Add a new index to `__reana.job` on `workflow_uuid` and `created` to
query efficiently the jobs that are part of a given workflow, with the
possibility of ordering them by creation time.

Add a new index to `__reana.workflow` on `status` to efficiently query
the total number of running workflows.

Change the already-existing index on `__reana.workflow` so that
`owner_id` is the leading column, before `name`.

Closes #211

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,7 +3,9 @@ Changes
 
 Version 0.9.3 (UNRELEASED)
 --------------------------
+
 - Changes the ``Workflow`` table to replace the ``run_number`` column with two new columns ``run_number_major`` and ``run_number_minor``, in order to allow for more than 9 restarts.
+- Changes the names of database constraints to follow the same naming convention.
 
 Version 0.9.2 (2023-09-26)
 --------------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,7 @@ Version 0.9.3 (UNRELEASED)
 
 - Changes the ``Workflow`` table to replace the ``run_number`` column with two new columns ``run_number_major`` and ``run_number_minor``, in order to allow for more than 9 restarts.
 - Changes the names of database constraints to follow the same naming convention.
+- Changes the database indexes to improve performance of common database queries.
 
 Version 0.9.2 (2023-09-26)
 --------------------------
--- a/reana_db/alembic/versions/20231128_1729_2461610e9698_enforce_naming_convention.py
+++ b/reana_db/alembic/versions/20231128_1729_2461610e9698_enforce_naming_convention.py
@@ -1,0 +1,213 @@
+"""Enforce naming convention.
+
+Revision ID: 2461610e9698
+Revises: b85c3e601de4
+Create Date: 2023-11-28 17:29:58.140440
+
+"""
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "2461610e9698"
+down_revision = "b85c3e601de4"
+branch_labels = None
+depends_on = None
+
+constraints = [
+    (
+        "workflow",
+        "workflow_pkey",
+        "pk_workflow",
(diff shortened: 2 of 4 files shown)


reana_db/alembic/versions/20231123_1342_126601b69c78_improve_indexes_usage.py
```diff
@@ -0,0 +1,71 @@
...
+        "_workflow_uuid_created_ix",
+        "job",
+        ["workflow_uuid", "created"],
+        unique=False,
+        schema="__reana",
+    )
+
+    # Create new index on (status) of __reana.workflow
+    op.create_index(
+        op.f("ix___reana_workflow_status"),
```
giuseppe-steduto, line 44: Minor: is there a naming convention to be consistent with the names of the index? See `_workflow_uuid_created_ix` and `ix___reana_workflow_status`, where `ix` is once used as a prefix, and once as a suffix. I see the one for the status is autogenerated by SQLAlchemy, maybe we can try to use the same?

giuseppe-steduto (approved):
Works well! Just left a minor cosmetic comments on the names of the indexes.

reana_db/alembic/versions/20231123_1342_126601b69c78_improve_indexes_usage.py
mdonadoni: Yes, SQLAlchemy autogenerates the name, but by default only for indexes. We could set up naming conventions also for foreign keys/etc. like this: https://docs.sqlalchemy.org/en/20/core/constraints.html#configuring-constraint-naming-conventions

However, we already have some inconsistencies:
```
Indexes:
    "workflow_pkey" PRIMARY KEY, btree (id_)
    "_user_workflow_run_uc" UNIQUE CONSTRAINT, btree (owner_id, name, run_number_major, run_number_minor)
    "ix___reana_workflow_status" btree (status)
Foreign-key constraints:
    "workflow_owner_id_fkey" FOREIGN KEY (owner_id) REFERENCES __reana.user_(id_)
Referenced by:
    TABLE "__reana.workflow_resource" CONSTRAINT "workflow_resource_workflow_id_fkey" FOREIGN KEY (workflow_id) REFERENCES __reana.workflow(id_)
    TABLE "__reana.workflow_session" CONSTRAINT "workflow_session_workflow_id_fkey" FOREIGN KEY (workflow_id) REFERENCES __reana.workflow(id_)
    TABLE "__reana.workspace_retention_rule" CONSTRAINT "workspace_retention_rule_workflow_id_fkey" FOREIGN KEY (workflow_id) REFERENCES __reana.workflow(id_)
```

- `workflow_pkey` and `workflow_owner_id_fkey` (generated by PostgreSQL)
- `_user_workflow_run_uc` (custom name that does not refer to any particular column)
- `ix___reana_workflow_status` (autogenerated by SQLAlchemy)

I can try to set some default naming conventions and see what happens, but I will need to update/create a migration to change the name of the already-existing constraints

mdonadoni: I have added a new migration to make all the names consistent between each other. Many changes, so lots of testing is needed ;)

mdonadoni:
Note that the chosen convention is the one present in SQLAlchemy's docs and which is used also by Invenio ([source](https://github.com/inveniosoftware/invenio-db/blob/76014ba16d47790f49738ea255525db236755151/invenio_db/shared.py#L17-L26))

PR #214, opened 2023-11-24
fix(models): add missing foreign key to workflow_uuid of Job (#214) 

### LINKED ISSUE #210: models: missing foreign key in Job table
There is a missing foreign key constraint for `workflow_uuid` in the Job table:
https://github.com/reanahub/reana-db/blob/589978bf621837ed1f0e02962f4793dacbae8461/reana_db/models.py#L817

Migration scripts also need to be prepared. Beware that in already existing instances of REANA there might be orphaned jobs (i.e. jobs without the related workflow) as in the past it was possible to hard-delete workflows (see https://github.com/reanahub/reana-client/issues/436).

mdonadoni:
Closes #210 

Depends on #213

### DIFF
--- a/reana_db/alembic/versions/20240531_0959_86435bb00714_foreign_key_for_workflow_uuid_of_job.py
+++ b/reana_db/alembic/versions/20240531_0959_86435bb00714_foreign_key_for_workflow_uuid_of_job.py
@@ -1,0 +1,102 @@
+"""Foreign key for workflow_uuid of Job.
+
+Revision ID: 86435bb00714
+Revises: eb5309f3d8ee
+Create Date: 2024-05-31 09:59:18.951074
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "86435bb00714"
+down_revision = "eb5309f3d8ee"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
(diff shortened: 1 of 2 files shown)


mdonadoni:
This PR does not handle the cleaning of the database to make sure that all the values in `workflow_uuid` are valid non-null foreign keys.

Should the cleanup be performed in the migration, by deleting the related records? Otherwise, we have to instruct the administrator to perform the cleanup manually

reana_db/alembic/versions/20240531_0959_86435bb00714_foreign_key_for_workflow_uuid_of_job.py
```diff
@@ -0,0 +1,64 @@
...
+    """Upgrade to 86435bb00714."""
+    job_table = sa.sql.table("job", sa.sql.column("workflow_uuid"), schema="__reana")
+    workflow_table = sa.sql.table("workflow", sa.sql.column("id_"), schema="__reana")
+
+    # 1. delete jobs which refer to non-existing workflows
+    op.execute(
+        job_table.delete().where(
+            job_table.c.workflow_uuid.notin_(sa.select([workflow_table.c.id_]))
+        )
+    )
```
mdonadoni, line 30: :warning:  Jobs that are not part of an existing workflows are deleted!

mdonadoni:
This was tested on a clone of DEV

tiborsimko (approved):

PR #216, opened 2023-12-15
utils: add permission error for getting workflows

DaanRosendal:
Closes reanahub/reana-server#651, closes reanahub/reana-server#650

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -4,6 +4,7 @@ The list of contributors in alphabetical order:
 
 - [Audrius Mecionis](https://orcid.org/0000-0002-3759-1663)
 - [Camila Diaz](https://orcid.org/0000-0001-5543-797X)
+- [Daan Rosendal](https://orcid.org/0000-0002-3447-9000)
 - [Diego Rodriguez](https://orcid.org/0000-0003-0649-2002)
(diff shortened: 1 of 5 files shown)


mdonadoni:
Latest version before rebase is a5726ff

Two things left:
- Fixing https://github.com/reanahub/reana-db/pull/212#pullrequestreview-1812979126 (maybe in a separate PR)
- Should we revert the last commit that adds `PermissionError`? As far as I can tell, these exceptions are never caught anywhere else in the codebase. Furthermore, the `PermissionError` exception is OS exception, so it should not be used here (see https://docs.python.org/3/library/exceptions.html#PermissionError and https://docs.python.org/3/library/exceptions.html#OSError)

mdonadoni:
Ready to be reviewed, squashed and merged

tiborsimko:
> Should we revert the last commit that adds `PermissionError`? 

Agreed, I'll remove it for now before merging.  (Kept local copy just in case.)

tiborsimko (approved):
Works nicely :+1: with

```
- r-client #692
- r-client-go #153
- r-ui #375
- r-server #658
- r-workflow-controller #552
- r-commons #429
- r-db #216
```

Removed the `PermissionError` commit and kept it locally.

PR #234, opened 2024-03-22
build(python): add support for SQLAlchemy 1.4 (#234)

### LINKED ISSUE #228: build(python): update SQLAlchemy to newer version
We are currently requiring SQLAlchemy<1.4.0:
https://github.com/reanahub/reana-db/blob/3fbcf65db735146d54078cae4c5b9c8968ead055/setup.py#L48

However, the last release of SQLAlchemy 1.3 was made long ago (March 31st 2021): https://pypi.org/project/SQLAlchemy/1.3.24/

SQLAlchemy needs to be updated at least to 1.4, and if possible to 2.0.

mdonadoni:
- build(python): add support for SQLAlchemy 1.4 (#234)
- fix(models): remove duplicated relationship (#234)

Closes #228

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2018, 2019, 2020, 2021, 2022, 2023 CERN.
+# Copyright (C) 2018, 2019, 2020, 2021, 2022, 2023, 2024 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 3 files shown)


mdonadoni:
New warnings:

_SAWarning: TypeDecorator EncryptedType() will not produce a cache key because the ``cache_ok`` attribute is not set to True.  This can have significant performance implications including some performance degradations in comparison to prior SQLAlchemy versions.  Set this attribute to True if this type object's state is safe to use in a cache key, or False to disable this warning. (Background on this error at: https://sqlalche.me/e/14/cprf)_

This can be partially fixed by changing `EncryptedType` (deprecated) into `StringEncryptedType` (note that a migration script would be needed), but some of the warnings are coming from some of invenio's dependencies.

I have not seen lower performance locally in any case.

tiborsimko (approved):
Works well (using lower boundary).

Please verify that the relationship backref change does not require any alembic upgrade recipe. (Wondering whether the index key names could have changed or some such?)

mdonadoni:
> Please verify that the relationship backref change does not require any alembic upgrade recipe. (Wondering whether the index key names could have changed or some such?)

 I have double checked this and autogenerating a new alembic recipe results in empty downgrade/upgrade methods. Indexes should not be affected by these changes.

PR #235, opened 2024-03-25
build(python): drop support for Python 3.6 and 3.7 (#235)

mdonadoni:
BREAKING CHANGE: drop support for Python 3.6 and 3.7

Closes reanahub/reana#784


.github/workflows/ci.yml
```diff
@@ -143,7 +143,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: ['3.8', '3.9']
```
mdonadoni, line 146: Which Python version should we test here?

I have done some tests with the `continue-on-error` setting, but while the workflow run does not fail, each job is still reported as failing on the PR checks, which I think is quite confusing. I think it's better to keep only passing Python versions.

mdonadoni: I have kept only Python 3.8 in `setup.py`'s classifiers and in the CI, here and in all the other non-shared components

PR #241, opened 2024-11-20
feat(models): add service tables

Alputer:
This PR adds `Service` table to db. Currently, we have only `dask` service but it is a generic implementation and can be used for other components associated with workflows in the future.

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -470,6 +470,63 @@ class InteractiveSession(Base, Timestamp, QuotaBase):
         return "<InteractiveSession %r>" % self.name
 
 
+class WorkflowService(Base):
+    """Workflow Service table."""
+
+    __tablename__ = "workflow_service"
+    __table_args__ = {"schema": "__reana"}
+
+    workflow_id = Column(UUIDType, ForeignKey("__reana.workflow.id_"), nullable=True)
+    service_id = Column(UUIDType, ForeignKey("__reana.service.id_"), primary_key=True)
+
+    def __repr__(self):
+        """Workflow Service string representation."""
+        return f"<WorkflowService {self.service_id} {self.workflow_id}>"
+
+
+class ServiceStatus(CleanUpDependingOnStatusMixin, enum.Enum):
+    """Enumeration of possible run statuses."""
+
+    created = 0
+    running = 1
+    finished = 2
+    failed = 3
+    deleted = 4
+    stopped = 5
+    queued = 6
+    pending = 7
+
+
+class ServiceType(enum.Enum):
+    """Enumeration of service types."""
+
+    dask = 0
+
+
+class Service(Base):
+    """Service table."""
+
+    __tablename__ = "service"
+    id_ = Column(UUIDType, primary_key=True, default=generate_uuid)
+    name = Column(String(255))
+    uri = Column(Text)  # uri to access the service
+    status = Column(Enum(ServiceStatus), nullable=False, default=ServiceStatus.created)
+    owner_id = Column(UUIDType, ForeignKey("__reana.user_.id_"))
+    type_ = Column(
+        Enum(ServiceType),
+        nullable=False,
+    )
+
+    __table_args__ = (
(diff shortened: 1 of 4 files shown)


reana_db/alembic/versions/20241119_1051_8da6ccf08af9_add_service_tables.py
```diff
@@ -0,0 +1,76 @@
+"""Service tables.
+
+Revision ID: 8da6ccf08af9
+Revises: 2e82f33ee37d
+Create Date: 2024-11-19 10:51:51.199670
+
```
tiborsimko, line 6: **todo:** ‏Remove empty line in the docstring.

reana_db/alembic/versions/20241119_1051_8da6ccf08af9_add_service_tables.py
```diff
@@ -0,0 +1,76 @@
...
+        ),
+        sa.ForeignKeyConstraint(
+            ["workflow_id"],
+            ["__reana.workflow.id_"],
+            name=op.f("fk_workflow_service_workflow_id_workflow"),
+        ),
+        sa.PrimaryKeyConstraint("service_id", name=op.f("pk_workflow_service")),
+        schema="__reana",
+    )
+    # ### end Alembic commands ###
```
tiborsimko, line 69: **todo:** ‏Remove the "end Alembic" comment.

reana_db/alembic/versions/20241119_1051_8da6ccf08af9_add_service_tables.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def downgrade():
+    """Downgrade to 2e82f33ee37d revision."""
+    op.drop_table("workflow_service", schema="__reana")
+    op.drop_table("service", schema="__reana")
+    # ### end Alembic commands ###
```
tiborsimko, line 76: **todo:** ‏Remove "end Alembic" comment.

reana_db/models.py
```diff
@@ -470,6 +470,50 @@ def __repr__(self):
...
+
+    def __repr__(self):
+        """Workflow Service string representation."""
+        return f"<WorkflowService {self.service_id} {self.workflow_id}>"
+
+
+class ServiceType(enum.Enum):
+    """Enumeration of service types."""
+
+    dask = 0
```
tiborsimko, line 490: Is this one intended to be general Dask service? I wonder about distinguishing Dask cluster from Dask dashboard, for example in case there would be more services exposed from the Dask integration. But I guess these theoretical distinctions may not be useful as of yet...

reana_db/models.py
```diff
@@ -470,6 +470,50 @@ def __repr__(self):
...
+
+
+class Service(Base):
+    """Service table."""
+
+    __tablename__ = "service"
+    id_ = Column(UUIDType, primary_key=True, default=generate_uuid)
+    name = Column(String(255))
+    uri = Column(Text)  # uri to access the service
+    status = Column(Enum(RunStatus), nullable=False, default=RunStatus.created)
```
tiborsimko, line 500: For admin use cases, it may be perhaps good to have some indexes, e.g. on the `status` column so that we could ask "which users have Dask dashboards still open?".  But this is something that we may also get to later.

reana_db/alembic/versions/20241119_1051_8da6ccf08af9_add_service_tables.py
Alputer: All other files (except workflow_sharing migration which has the additional rebase_date field) have the same format. Maybe better to follow the same structure? The empty line might be signaling that no rebasing has taken place.

reana_db/models.py
Alputer: I think we can introduce a new service type relatively easily if we need an extra field in the future. Calling the service `dask_dashboard` or something similar at the moment could be too specific.

tiborsimko (approved):

PR #248, opened 2025-05-27
feat(models): add ServiceLogs table

Alputer:


reana_db/models.py
```diff
@@ -503,6 +504,26 @@ class ServiceType(enum.Enum):
     dask = 0
 
 
+class ServiceLog(Base):
```
tiborsimko, line 507: **todo:** ‏We'd need an alembic recipe for the new table.

Alputer: Added

PR #250, opened 2025-07-21
feat: add identity provider ID to User model

tomondre:
* Add `idp_issuer` and `idp_subject` columns to User model for external identity provider integration
* Support mapping external identity provider subject claims to REANA users
* Enable user lookup by identity provider ID

This change allows storing external identity provider subject identifiers in the database, supporting the JWT authentication feature by creating a mapping between external identity providers and REANA user accounts.

More information on our efforts and progress can be found in [this issue](https://github.com/reanahub/reana-server/issues/743)

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -178,6 +178,8 @@ class User(Base, Timestamp, QuotaBase):
         sync_backref=False,
     )
     audit_logs = relationship("AuditLog", backref="user_")
+    idp_issuer = Column(String(length=255))
+    idp_subject = Column(String(length=255))
 
     def __init__(self, access_token=None, **kwargs):
         """Initialize user model."""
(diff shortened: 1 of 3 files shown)


reana_db/models.py
```diff
@@ -178,6 +178,7 @@ class User(Base, Timestamp, QuotaBase):
         sync_backref=False,
     )
     audit_logs = relationship("AuditLog", backref="user_")
+    idp_id = Column(String(length=255))
```
tiborsimko, line 181: Please note that changing the data model would require us to add DB schema migration scripts. We use Alembic for this. Here is the recipe how to generate one: https://github.com/reanahub/reana/wiki/Tips-for-SQLAlchemy#autogenerate-an-alembic-migration

tiborsimko, line 181: I hope that the IdP ID limit of max 255 characters will suffice in general. Should be the case, but can you please quick-check the various popular AAI systems out there in order to be fully sure?

tiborsimko:
Closing; the move to OIDC and JWT was completed recently in reanahub/reana#977 and its related PRs.

PR #255, opened 2026-03-20
feat(models): add quota period database support

michaelbuchar:
Closes https://github.com/reanahub/reana-server/issues/774

### DIFF
--- a/reana_db/cli.py
+++ b/reana_db/cli.py
@@ -16,11 +16,16 @@ import click
 from alembic import command
 from alembic import config as alembic_config
 from reana_commons.config import REANA_LOG_FORMAT, REANA_LOG_LEVEL
+from reana_db.config import (
+    DEFAULT_QUOTA_CPU_PERIOD_RESET_MONTHS,
+)
 
 
 from reana_db.database import init_db
 from reana_db.models import Resource, ResourceType
 from reana_db.utils import (
+    get_current_quota_period_start_at,
+    get_default_quota_resource,
     change_key_encrypted_columns,
     update_users_cpu_quota,
     update_users_disk_quota,
@@ -272,15 +277,15 @@ def resource_usage_update() -> None:
         """Update users resource quota usage."""
         try:
             if resource == ResourceType.disk:
-                update_workflows_disk_quota()
-                update_users_disk_quota()
+                update_workflows_disk_quota(override_policy_checks=True)
+                update_users_disk_quota(override_policy_checks=True)
                 click.secho(
                     "Disk quota usage updated successfully for all users and workflows.",
                     fg="green",
                 )
             elif resource == ResourceType.cpu:
-                update_workflows_cpu_quota()
-                update_users_cpu_quota()
+                update_workflows_cpu_quota(override_policy_checks=True)
+                update_users_cpu_quota(override_policy_checks=True)
                 click.secho(
                     "CPU quota usage updated successfully for all users and workflows.",
                     fg="green",
@@ -294,3 +299,50 @@ def resource_usage_update() -> None:
 
     _resource_usage_update(ResourceType.disk)
     _resource_usage_update(ResourceType.cpu)
+
+
+@quota_group.command("init-default-period")
+@click.option("--yes", is_flag=True, help="Do not ask for confirmation.")
+@click.option("--dry-run", is_flag=True, help="Report what would change.")
+def init_default_period(yes: bool, dry_run: bool) -> None:
+    """Backfill deployment default periodic CPU quota policy for existing users."""
+    from reana_db.database import Session
+    from reana_db.models import UserResource
+
+    if not DEFAULT_QUOTA_CPU_PERIOD_RESET_MONTHS:
+        click.secho(
+            "Default CPU quota period is not configured.",
+            fg="yellow",
+        )
+        return
+
+    cpu_resource = get_default_quota_resource(ResourceType.cpu.name)
+    user_resources = (
+        Session.query(UserResource)
+        .filter(UserResource.resource_id == cpu_resource.id_)
+        .filter(UserResource.quota_period_months.is_(None))
+        .all()
+    )
+    updated_rows = len(user_resources)
+
+    click.echo(
+        f"Will set the default CPU quota period for {updated_rows} users "
+        f"({DEFAULT_QUOTA_CPU_PERIOD_RESET_MONTHS} months)."
+    )
+    if dry_run:
+        return
+    if not yes:
+        click.confirm("Proceed?", abort=True)
+
+    for user_resource in user_resources:
+        user_resource.quota_period_months = DEFAULT_QUOTA_CPU_PERIOD_RESET_MONTHS
+        user_resource.quota_period_start_at = get_current_quota_period_start_at(
+            reference_start_at=user_resource.user.created,
+            quota_period_months=DEFAULT_QUOTA_CPU_PERIOD_RESET_MONTHS,
+        )
+    Session.commit()
+
+    click.secho(
+        f"Initialised the default CPU quota period for {updated_rows} existing users.",
+        fg="green",
+    )
--- a/reana_db/config.py
+++ b/reana_db/config.py
@@ -76,6 +76,27 @@ DEFAULT_QUOTA_LIMITS = {
 """Default CPU (in milliseconds) and disk (in bytes) quota limits."""
 
 
+def _get_optional_period_months_env(var_name):
+    value = os.getenv(var_name)
+    if value in (None, ""):
+        return None
+
+    try:
+        parsed = int(value)
+    except ValueError as exc:
+        raise ValueError(f"{var_name} must be a non-negative integer") from exc
+
+    if parsed < 0:
+        raise ValueError(f"{var_name} must be a non-negative integer")
+
+    return parsed or None
+
+
+DEFAULT_QUOTA_CPU_PERIOD_RESET_MONTHS = _get_optional_period_months_env(
+    "REANA_DEFAULT_QUOTA_CPU_PERIOD_RESET_MONTHS"
+)
+"""Length of the default CPU accounting window in months for newly created users."""
+
 policies = os.getenv("REANA_WORKFLOW_TERMINATION_QUOTA_UPDATE_POLICY")
 WORKFLOW_TERMINATION_QUOTA_UPDATE_POLICY = policies.split(",") if policies else []
 """What quota types to update, if not specified all quotas will be calculated, if empty no quotas will be updated."""
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -30,11 +30,13 @@ from reana_commons.utils import get_disk_usage
 import reana_db.config
 from reana_db.config import (
     DEFAULT_QUOTA_LIMITS,
+    DEFAULT_QUOTA_CPU_PERIOD_RESET_MONTHS,
     DEFAULT_QUOTA_RESOURCES,
     WORKFLOW_TERMINATION_QUOTA_UPDATE_POLICY,
 )
 from reana_db.utils import (
     build_workspace_path,
+    get_current_quota_period_start_at,
     store_workflow_disk_quota,
     split_run_number,
     update_users_cpu_quota,
@@ -45,6 +47,7 @@ from reana_db.utils import (
 from sqlalchemy import (
     BigInteger,
     Boolean,
+    CheckConstraint,
     Column,
     DateTime,
     Enum,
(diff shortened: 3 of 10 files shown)


reana_db/cli.py
```diff
@@ -279,15 +278,15 @@ def _resource_usage_update(resource: ResourceType) -> None:
         """Update users resource quota usage."""
         try:
             if resource == ResourceType.disk:
-                update_workflows_disk_quota()
-                update_users_disk_quota()
+                update_workflows_disk_quota(override_policy_checks=True)
```
michaelbuchar, line 281: Let me know if this is acceptable, I thought you should be able to override policy checks if you want to manually re-trigger the quota `resource-usage-update`

reana_db/alembic/versions/20260414_1535_8f2d4e5a4d2c_drop_user_resource_quota_period_anchor.py
```diff
@@ -0,0 +1,84 @@
...
+    column_names = {
+        column["name"]
+        for column in inspector.get_columns("user_resource", schema="__reana")
+    }
+    if "quota_period_anchor_at" in column_names:
+        op.drop_column("user_resource", "quota_period_anchor_at", schema="__reana")
+
+
+def downgrade():
+    """Keep the schema introduced in revision 06dbbeef6d9b."""
```
tiborsimko: 🤖 **Empty `downgrade()` makes this migration non-reversible (BLOCKER)**

`upgrade()` drops `quota_period_anchor_at` (line 80), but `downgrade()` is a no-op. If a deployment needs to roll back to the prior schema after spotting a regression, Alembic will report success without restoring the column, and the previous code path that referenced `quota_period_anchor_at` will fail at the next read.

Suggested fix — recreate the column (the original values cannot be recovered, but at least the schema is restored):

```python
def downgrade():
    """Recreate the legacy anchor column (will be NULL for all rows)."""
    op.add_column(
        "user_resource",
        sa.Column("quota_period_anchor_at", sa.DateTime(), nullable=True),
        schema="__reana",
    )
```

tiborsimko (commented):
Inline review of the quota period changes. Posting a punch list of items grouped by severity (BLOCKER / HIGH / MEDIUM / LOW / NIT). Happy to discuss any of these.

reana_db/alembic/versions/20260414_1535_8f2d4e5a4d2c_drop_user_resource_quota_period_anchor.py
```diff
@@ -0,0 +1,84 @@
...
+                SET quota_period_start_at = :quota_period_start_at
+                WHERE user_id = :user_id
+                  AND resource_id = :resource_id
+                """),
+            {
+                "quota_period_start_at": current_period_start_at,
+                "user_id": row["user_id"],
+                "resource_id": row["resource_id"],
+            },
+        )
```
tiborsimko: 🤖 **Per-row `UPDATE` loop will be slow on populated production databases (HIGH)**

The migration issues one `UPDATE` round-trip per affected row. For a deployment with thousands of users (one row per resource per user), this can take a noticeable amount of time and extends the window during which the table is being mutated. The actual computation is a deterministic function of `account_created_at`, `quota_period_months`, and `now()`, so it can be folded into a single SQL statement.

Suggested fix — a pure-SQL `UPDATE ... FROM` using a recursive CTE that walks `account_created_at` forward by `quota_period_months` until it reaches the current window. If we prefer to keep the Python loop for readability, batch the writes into a single `connection.execute(text(...), [...])` call so we make one round-trip instead of N.

If we don't expect populated production databases to have many periodic-quota rows at the time this migration runs, please add a comment in the migration acknowledging that assumption.

reana_db/utils.py
```diff
@@ -491,7 +493,109 @@ def update_workflow_cpu_quota(workflow) -> int:
...
+    if original_period_start_at == current_period_start_at:
+        return False
+
+    user_resource_quota.quota_period_start_at = current_period_start_at
+    logging.info(
+        "Set current CPU quota period for user %s from %s to %s",
+        user_resource_quota.user_id,
+        original_period_start_at,
+        current_period_start_at,
+    )
```
tiborsimko, line 594: 🤖 **`logging.info` on every period advance will spam logs at scale (MEDIUM)**

`update_users_cpu_quota` runs from a periodic cron job over all users with active tokens. Whenever a period rolls over, this INFO line is emitted. With many users and a short cadence, the signal-to-noise ratio is low — the change is also visible in the database as `quota_period_start_at`, so operators don't need a per-user log entry.

Suggested fix — demote to `logging.debug`:

```python
logging.debug(
    "Set current CPU quota period for user %s from %s to %s",
    ...
)
```

reana_db/cli.py
```diff
@@ -294,3 +300,44 @@ def _resource_usage_update(resource: ResourceType) -> None:
...
+                quota_period_months=DEFAULT_QUOTA_CPU_PERIODIC_RESET_MONTHS,
+            )
+        )
+    Session.commit()
+    updated_rows = len(user_resources)
+
+    click.secho(
+        f"Initialised periodic CPU reset policy for {updated_rows} existing users.",
+        fg="green",
+    )
```
tiborsimko: 🤖 **`init-periodic-defaults` is an unguarded bulk write (MEDIUM)**

The command updates every CPU `UserResource` whose `quota_period_months` is NULL — potentially every existing user — and commits in one go. There is no dry-run preview, no confirmation, and no count printed before the write. Running it on the wrong cluster, or with the wrong default months value in the environment, would be hard to undo.

Suggested fix — add a `--yes` / `--dry-run` pair and print the affected count before committing:

```python
@quota_group.command("init-periodic-defaults")
@click.option("--yes", is_flag=True, help="Do not ask for confirmation.")
@click.option("--dry-run", is_flag=True, help="Report what would change.")
def init_periodic_defaults(yes: bool, dry_run: bool) -> None:
    ...
    click.echo(
        f"Will set periodic CPU quota for {len(user_resources)} users "
        f"({DEFAULT_QUOTA_CPU_PERIODIC_RESET_MONTHS} months)."
    )
    if dry_run:
        return
    if not yes:
        click.confirm("Proceed?", abort=True)
    ...
```

reana_db/alembic/versions/20260414_1535_8f2d4e5a4d2c_drop_user_resource_quota_period_anchor.py
```diff
@@ -0,0 +1,84 @@
...
+def _get_current_quota_period_start_at(account_created_at, quota_period_months):
+    """Derive the active quota window start from the account creation time."""
+    if not account_created_at or not quota_period_months:
+        return None
+
+    now = datetime.utcnow()
+    period_start_at = account_created_at
+    while now >= _add_months(period_start_at, quota_period_months):
+        period_start_at = _add_months(period_start_at, quota_period_months)
+    return period_start_at
```
tiborsimko: 🤖 **Period-arithmetic helpers are duplicated between the migration and `utils.py` (LOW)**

`_add_months` and `_get_current_quota_period_start_at` are byte-identical to the equivalents in `reana_db/utils.py:520-541`. Self-containment is the right choice for migrations (they should not import from app code that may evolve), but it does mean a future bug fix to the period arithmetic — for example, the day-clamping at month boundaries — needs to land in two places.

Suggested fix — add a short comment in both copies pointing at the other so this is hard to miss during future maintenance. For example, above the migration helper:

```python
# NOTE: kept in sync with reana_db.utils._add_months /
# get_current_quota_period_start_at_from_anchor. Migrations must be
# self-contained, so any change to the period arithmetic must be applied
# in both places.
```

reana_db/models.py
```diff
@@ -1266,6 +1286,9 @@ class UserResource(Base, Timestamp):
     resource_id = Column(UUIDType, ForeignKey("__reana.resource.id_"), primary_key=True)
     quota_limit = Column(BigInteger())
     quota_used = Column(BigInteger())
+    quota_period_months = Column(Integer, nullable=True)
```
tiborsimko, line 1294: 🤖 **No `CHECK` constraint on `quota_period_months` (LOW)**

The column accepts any integer at the database level, including 0 and negatives. Validation is left entirely to the application layer (`reana_server`). A `CheckConstraint` makes the invariant explicit and protects against manual SQL updates or future callers that bypass the API.

Suggested fix:

```python
quota_period_months = Column(
    Integer,
    CheckConstraint(
        "quota_period_months IS NULL OR quota_period_months > 0",
        name="ck_quota_period_months_positive",
    ),
    nullable=True,
)
```

Note: this also requires adding the constraint in migration `06dbbeef6d9b` (or a follow-up migration), and removing it in `downgrade()`.

tests/test_utils.py
```diff
@@ -91,3 +107,202 @@ def test_get_workflow_with_uuid_or_name(session, new_user):
...
+        (None, 3, datetime(2026, 4, 14, 0, 0, 0), None),
+        (
+            datetime(2026, 1, 15, 12, 0, 0),
+            3,
+            datetime(2026, 5, 20, 0, 0, 0),
+            datetime(2026, 4, 15, 12, 0, 0),
+        ),
+    ],
+)
+def test_get_current_quota_period_start_at_from_anchor(
```
tiborsimko: 🤖 **Missing tests for month-boundary day clamping (NIT)**

`_add_months` (utils.py:525) clamps the day to the last valid day of the target month. This means a user anchored at Jan 31 with a 1-month period gets boundaries Jan 31 → Feb 28 → Mar 28 → ... rather than rolling back to the 31st in months that have 31 days. That is a real semantic choice and the tests should pin it down so a future "simplification" doesn't silently change behaviour.

Suggested fix — add parameterised cases for `_add_months` over `(2026-01-31, 1)`, `(2024-02-29, 12)` (leap year → 2025-02-28), `(2026-12-31, 2)`, and a `get_current_quota_period_start_at_from_anchor` test for a Jan-31 anchor advanced 14 months.

reana_db/utils.py
```diff
@@ -549,20 +658,42 @@ def update_users_cpu_quota(user=None) -> None:
...
+            continue
+
+        _advance_user_cpu_quota_period_if_needed(user_resource_quota)
+
+        quota_period_start_at = _get_current_user_cpu_quota_period_start_at(
+            user_resource_quota
+        )
+
+        if quota_period_start_at:
+            workflows = Session.query(Workflow).filter_by(owner_id=user.id_).all()
```
tiborsimko: 🤖 **Periodic CPU recomputation loads all workflows per user (HIGH)**

For users with periodic CPU accounting enabled, this path loads every `Workflow` row for the owner and then sums clipped CPU time in Python. On populated deployments this can become expensive: the query runs once per user, and unlike `update_workflows_cpu_quota()` just above, it does not defer heavy columns such as logs or workflow specifications. The periodic cron can therefore become much slower and more memory-hungry exactly when it processes users with many historical workflows.

Suggested fix — avoid loading full workflow objects here. At minimum, select only the timestamp columns needed for `_get_accounted_workflow_cpu_milliseconds` or apply the same `defer(Workflow.logs), defer(Workflow.reana_specification)` pattern used in `update_workflows_cpu_quota()`. Longer term, consider calculating the clipped sum in SQL so the periodic path remains one aggregate query per user instead of materialising all workflows in Python.

reana_db/config.py
```diff
@@ -76,6 +76,21 @@
 """Default CPU (in milliseconds) and disk (in bytes) quota limits."""
 
 
+def _get_optional_int_env(var_name):
+    value = os.getenv(var_name)
+    return int(float(value)) if value not in (None, "") else None
```
tiborsimko: 🤖 **Default period months parsing accepts surprising values (MEDIUM)**

`_get_optional_int_env()` parses through `float` before converting to `int`. This means values such as `"3.9"` silently become `3`, while `0` or negative values are accepted at config-load time and later disappear through truthiness checks. For an operator-facing Helm/environment setting that controls quota accounting cadence, silently coercing invalid values makes misconfiguration hard to diagnose.

Suggested fix — parse this as a strict positive integer and fail early for invalid values:

```python
def _get_optional_positive_int_env(var_name):
    value = os.getenv(var_name)
    if value in (None, ""):
        return None
    parsed = int(value)
    if parsed <= 0:
        raise ValueError(f"{var_name} must be a positive integer")
    return parsed
```

This also avoids accepting decimal strings whose truncation would be unexpected.

reana_db/alembic/versions/20260414_1535_8f2d4e5a4d2c_drop_user_resource_quota_period_anchor.py
michaelbuchar: Fixed!

reana_db/alembic/versions/20260414_1535_8f2d4e5a4d2c_drop_user_resource_quota_period_anchor.py
michaelbuchar: I think it's good to do this optimization, I have implemented this as well!

reana_db/utils.py
michaelbuchar: I think we should keep this at logging.info since it's good to transmit these messages but let me know if you disagree

reana_db/cli.py
michaelbuchar: Fixed!

reana_db/alembic/versions/20260414_1535_8f2d4e5a4d2c_drop_user_resource_quota_period_anchor.py
michaelbuchar: The duplication is intentional because the migration must remain self-contained and not depend on runtime application code.

reana_db/models.py
michaelbuchar: Fixed!

tests/test_utils.py
michaelbuchar: Fixed!

reana_db/utils.py
michaelbuchar: Fixed

reana_db/config.py
michaelbuchar: Fixed

tiborsimko (approved):

PR #257, opened 2026-04-26
build(python): upgrade to SQLAlchemy 2.x

### LINKED ISSUE #770: build(python): upgrade to latest Invenio packages, Flask 3.x, SQLAlchemy 2.x
Upgrade the Invenio dependency stack to latest versions:

- Invenio packages: invenio-db 2.5.0, invenio-accounts 7.x, invenio-app 3.x, invenio-oauthclient 7.x (upstream, replacing custom fork), and all other invenio-* packages to their latest major versions
- Flask 2.x → 3.x
- SQLAlchemy 1.4 → 2.x (via invenio-db 2.5.0 and reana-db)
- marshmallow 2.x → 3.x (required by new invenio-rest)
- webargs 5.x/6.x → 8.x
- Remove custom invenio-oauthclient fork (upstream now includes EOSC EU Node AAI provider)

This is a cross-component change affecting reana-db, pytest-reana, reana-commons, reana-server, reana-workflow-controller, reana-job-controller, reana-ui, and reana (Helm templates).

tiborsimko:
Upgrade SQLAlchemy from `<1.5.0` to `>=2.0.0` and SQLAlchemy-Utils to `>=0.41.0,<0.42.0` to support the latest Invenio packages and to unblock Snakemake `>=9.17.0` which depends on sqlmodel (requires SQLAlchemy 2.x).

Adapt the database session factory to the new API by removing the `autocommit` parameter from `sessionmaker()` (which is gone in 2.x) and removing `Base.query = Session.query_property()` (since the `Model.query` shortcut is no longer supported). Replace all `Model.query` call sites in models and utility code with explicit `Session.query(Model)` queries.

Adjust the connection-based code paths to the SQLAlchemy 2.x API. The `init_db()` schema check is now wrapped in `engine.connect()` since `dialect.has_schema()` requires a connection rather than an engine. The workspace retention event listener uses `.values()` when executing inserts, since keyword-style execution is no longer supported.

Fix the `declarative_base` import path, which moved from `sqlalchemy.ext.declarative` to `sqlalchemy.orm`. Update one Alembic migration to use the keyword-style `sa.select(column)` syntax instead of the deprecated list-based `sa.select([column])`.

Closes reanahub/reana-server#770.

### DIFF
--- a/reana_db/models.py
+++ b/reana_db/models.py
@@ -60,7 +60,7 @@ from sqlalchemy import (
     or_,
     text,
 )
-from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import declarative_base
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import relationship, backref
 from sqlalchemy_utils import EncryptedType, JSONType, UUIDType
@@ -282,7 +282,9 @@ class User(Base, Timestamp, QuotaBase):
 
     def initialize_user_quota_limits(self):
         """Initialize user quota limits."""
-        resources = Resource.query.all()
+        from .database import Session
+
+        resources = Session.query(Resource).all()
         for resource in resources:
             self.resources.append(
                 UserResource(
@@ -770,7 +772,9 @@ class Workflow(Base, Timestamp, QuotaBase):
 
     def activate_workspace_retention_rules(self):
         """Activate workspace retention rules for the workflow."""
-        rules = WorkspaceRetentionRule.query.filter_by(
+        from .database import Session
+
+        rules = Session.query(WorkspaceRetentionRule).filter_by(
             workflow_id=self.id_, status=WorkspaceRetentionRuleStatus.created
         )
         update_workspace_retention_rules(rules, WorkspaceRetentionRuleStatus.active)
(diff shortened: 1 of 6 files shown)


reana_db/models.py
```diff
@@ -60,14 +60,20 @@
...
+from sqlalchemy.orm import declarative_base
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import relationship, backref
 from sqlalchemy_utils import EncryptedType, JSONType, UUIDType
 from sqlalchemy_utils.models import Timestamp
 from sqlalchemy_utils.types.encrypted.encrypted_type import AesEngine
 from sqlalchemy.dialects.postgresql import ARRAY, JSONB
 
+
+class CachedEncryptedType(EncryptedType):
```
michaelbuchar: ## AI Review 
### CachedEncryptedType currently breaks the docs build

This helper class makes the docs-sphinx job fail on the PR. Sphinx tries to render the inherited cache_ok attribute docs for CachedEncryptedType and hits unresolved SQLAlchemy cross-references, so the docs build stays red even though the runtime test matrix passes. Please either hide this helper from autodoc, override the attribute docs, or otherwise adjust the docs config so the PR can merge cleanly.

tiborsimko: Good catch, thanks. On closer inspection, `CachedEncryptedType` turned out to be redundant: starting with `sqlalchemy-utils >= 0.41.0` (which `setup.py` already pins), `StringEncryptedType` — the parent of `EncryptedType` — sets `cache_ok = True`, so the SQLAlchemy 2.x caching warning is already silenced by inheritance. I've dropped the helper class and switched the column back to `EncryptedType` directly, which removes the docs build issue at the source. No Alembic migration needed since the column DDL is unchanged.

I have also fixed the `black` and `flake8` CI findings.

michaelbuchar (approved):

PR #265, opened 2026-05-26
feat(models): add uses_kubernetes column to Workflow table

CameronMcClymont:
The column is True if any steps in the workflow run on Kubernetes.


CameronMcClymont:
Not sure how the revision ID for the database migration is normally generated, so the current one is a placeholder.

CameronMcClymont:
Closing due to alternative implementation in https://github.com/reanahub/reana-server/pull/780 meaning this PR is no longer needed.

PR #267, opened 2026-06-12
feat(models): add compute_backends column to Workflow table

CameronMcClymont:
- Add a new `compute_backends` list[string] column representing which backends the workflow uses (e.g. kubernetes, ht-condor, etc)
- Adapt the `get_workflow_overload_priority` function to return the worst per-resource ratio across all compute-backend caps and the Dask cap
- Add a new `count_active_per_backend` function to count the pending/running workflows per concurrency-capped resource

Closes reanahub/reana-workflow-controller#663


michaelbuchar (dismissed):

reana_db/models.py
```diff
@@ -608,6 +608,7 @@ class Workflow(Base, Timestamp, QuotaBase):
     job_progress = Column(JSONType, default=dict)
     workspace_path = Column(String)
     restart = Column(Boolean, default=False)
+    uses_kubernetes = Column(Boolean, default=True, nullable=False)
```
tiborsimko: **🤖 RWC663-03: `uses_kubernetes` boolean is the wrong shape, and the surrounding data model already says more than it persists (MEDIUM)**

The persisted column is named after one specific scheduler question ("does this take a slot in the k8s concurrent-batch cap?") rather than after the underlying facts about a workflow. That naming choice has two costs and an interesting third observation about what is and isn't already in the model.

`reana-db/reana_db/models.py:611`:

```python
uses_kubernetes = Column(Boolean, default=True, nullable=False)
```

**First, the boolean collapses information the codebase actually needs.** REANA workflows are routinely hybrid: a single DAG can have some steps on Kubernetes and others on HTCondor or Slurm. The new `uses_kubernetes()` classifier handles that correctly in the "does any step touch k8s?" sense — `complexity.py:91-94` even calls this out:

> *"…this function scans all steps so that hybrid workflows whose first steps are on an external backend but whose later steps run on Kubernetes are not incorrectly treated as external-only."*

But the column it writes into is one bit, so any subsequent code that wants to know "which backends does this workflow use?" — for reporting, for accounting, for a future per-backend cap, for UI filters like "show me my HTCondor runs" — has to either re-parse `reana_specification` (slow, fragile) or join `Job.compute_backend` after submission (only works post-hoc). Today there is no DB-level way to ask "which backends did this workflow use?" at all.

**Second, the boolean's name does not generalise to new questions.** If HTCondor or Slurm grows a per-tenant cap tomorrow, or a new backend appears, the data model needs another column and another alembic migration each time. The same is true for future workflow-level Kubernetes resources beyond Dask (virtual-kubelet, GPU pools, future workflow-level services).

**Third — and this is the unexpected finding — much of what we'd want is already in the model.** The previous iteration of this finding suggested adding a `kubernetes_services` column. That name would collide head-on with what already exists. `Workflow.services` is a many-to-many relation to a `Service` model, and that relation today holds exactly the kind of object we'd want to record: `reana-server/reana_server/utils.py:820` builds a `Service(type_=ServiceType.dask, …)` row and `ensure_dask_service()` (line 830) appends it to `workflow.services` whenever the spec declares `workflow.resources.dask`. So:

- "Is this workflow using Dask?" → `any(s.type_ == ServiceType.dask for s in workflow.services)` — no new column needed.
- "What workflow engine?" → `Workflow.type_` (already a column, `String(30)`).
- "Which compute backends does the DAG use?" → **not in the data model anywhere**, only in the JSON `reana_specification` and in per-job `Job.compute_backend` rows after submission.

The gap is just the third one.

**Suggested fix — fill the missing gap, lean on what already exists.** Replace the boolean with a single `compute_backends` ARRAY that captures the set of execution backends the DAG uses:

```python
compute_backends = Column(
    ARRAY(String),
    nullable=False,
    default=lambda: ["kubernetes"],  # callable, not literal list — avoids the shared-
                                     # mutable-default footgun even though SQLAlchemy
                                     # copies on insert.
)
# Set of step-level execution backends declared in the workflow spec,
# drawn from REANA_COMPUTE_BACKENDS. Populated at submission time.
# Hybrid workflows naturally carry multiple entries, e.g. ["kubernetes", "htcondor"].
```

In the matching Alembic migration, set the corresponding server-side default the same way the `uses_kubernetes` migration sets `server_default=sa.text("true")` — for an ARRAY column, `server_default=sa.text("ARRAY['kubernetes']")` — so that the existing rows backfilled on upgrade do not depend on Python-side defaults at insert time.

The scheduler's "does this take a slot in the k8s cap?" question becomes the OR of:

1. `"kubernetes" in workflow.compute_backends` — the step-tree case.
2. `any(s.type_ == ServiceType.dask for s in workflow.services)` — the auxiliary-k8s-service case, using the relation that's already populated.

This puts every fact in exactly one place, models hybrid workflows natively, gives reporting and UI direct column/relation access, and lets new workflow-level k8s services slot into the existing `Service` model rather than triggering further Workflow-column migrations. The `Workflow.type_` column already covers the workflow-engine dimension; nothing new is needed there.

If a finer-grained weighted cap (per [finding RWC663-02](https://github.com/reanahub/reana-server/pull/780#discussion_r3421401195)) is adopted later, the natural next step is a typed JSON `resource_profile` carrying per-resource size/weight (Dask worker count, requested memory, GPU asks, etc.), to be summed against the cap. But that is a follow-up — `compute_backends` plus the existing `services` relation closes the immediate modelling gap.

If the boolean stays, at minimum the column docstring and the classifier comment should call out every workflow-level k8s-spawning resource that needs to be inspected, and a unit test should fail loudly when a new resource type is added without updating the classifier.

CameronMcClymont:
Some CI tests failing due to dependency on reana-commons PR: https://github.com/reanahub/reana-commons/pull/541

reana_db/models.py
CameronMcClymont: Replaced the `uses_kubernetes` column with a more flexible and informative `compute_backends` column.

CameronMcClymont:
Pushed one follow-up here as part of [RWC663-06](https://github.com/reanahub/reana-commons/pull/541#discussion_r3611447985), which flagged that a concurrency cap of `0` meant three different things across the stack.

`get_workflow_overload_priority()` skipped zero caps as falsy, so a user holding workflows on a backend capped at `0` kept full priority even though the scheduler would admit nothing for them. A zero cap now counts as saturated when the user has any active workflow on that resource, and is ignored when they have none — so the priority calculation and the scheduler see the same population. Covered by `test_overload_priority_with_closed_backend`.

The matching changes are cap validation in reanahub/reana-commons#541 and 0%-available dashboard reporting in reanahub/reana-server#780.

PR #269, opened 2026-06-19
auth rework

### LINKED ISSUE #977: Improve SSO and use JWT for external authentication
Improve SSO allow only one provider and use JWT for external authentication.

Sohn123:
Close: reanahub/reana#977


reana_db/models.py
```diff
@@ -163,12 +163,22 @@ class User(Base, Timestamp, QuotaBase):
     """User table."""
 
     __tablename__ = "user_"
-    __table_args__ = {"schema": "__reana"}
+    __table_args__ = (
+        UniqueConstraint("idp_issuer", "idp_subject", name="uq_user__idp_identity"),
```
tiborsimko, line 162: **🤖 PR269-02: New persistence invariants have no direct model tests (LOW)**

The branch adds the `(idp_issuer, idp_subject)` identity constraint and two encrypted secret columns (`gitlab_webhook_secret` and `InteractiveSession.session_secret`), but its only new test covers the quota-init CLI. Downstream server and controller tests exercise the happy-path use of these fields, but none directly verifies the database behaviours on which those components depend:

- inserting two users with the same issuer and subject must fail;
- encrypted values must survive commit, eviction from the SQLAlchemy identity map, reload, and decryption.

Without a direct persistence test, a model/type drift can surface later as an authentication or interactive-session failure in another component.

Suggested fix — add focused model tests for an identity collision and for store/commit/expunge/reload round trips of both encrypted columns.

reana_db/models.py
```diff
@@ -163,12 +163,22 @@ class User(Base, Timestamp, QuotaBase):
...
+        {"schema": "__reana"},
+    )
 
     id_ = Column(UUIDType, primary_key=True, unique=True, default=generate_uuid)
     email = Column(String(length=255), unique=True, primary_key=True)
     full_name = Column(String(length=255))
     username = Column(String(length=255))
+    idp_issuer = Column(String(length=255))
+    idp_subject = Column(String(length=255))
```
tiborsimko, line 180: **🤖 PR269-03: Nullable identity columns permit partial `(issuer, NULL)` identities (LOW)**

Both identity columns are nullable and only jointly unique (`reana_db/models.py:167,175-176`):

```python
UniqueConstraint("idp_issuer", "idp_subject", name="uq_user__idp_identity"),
...
idp_issuer = Column(String(length=255))
idp_subject = Column(String(length=255))
```

In PostgreSQL, rows containing a NULL never collide under a unique constraint, so any number of users may share `(issuer, NULL)` or `(NULL, subject)`. Nullable columns are right for the additive migration (existing rows must stay valid), but a half-set identity is always a bug — a user visible to issuer-scoped queries yet unmatchable by subject lookup — and nothing prevents application code from writing one.

Suggested fix — add `CheckConstraint("(idp_issuer IS NULL) = (idp_subject IS NULL)", name="ck_user__idp_identity_pair")` alongside the unique constraint (with the matching Alembic revision), so identities are atomically both-set or both-NULL.

reana_db/cli.py
```diff
@@ -266,7 +266,6 @@ def create_default_resources():
         click.secho(
             "No action to be taken: default resources already exist.", fg="yellow"
```
tiborsimko, line 280: **🤖 PR269-04: Default-resource initialisation is idempotent but not concurrent-safe (LOW)**

The `create-default-resources` command now exits successfully when resources already exist (`reana_db/cli.py:259-267`), which fixes sequential Helm re-runs. `Resource.initialise_default_resources()` remains a check-then-insert operation (`reana_db/models.py:1264-1288`): two initialisers can both observe a missing name, and the loser then receives an uncaught `IntegrityError` from the unique resource-name constraint.

A retried Helm hook overlapping a manual invocation can therefore still produce the deployment failure that this change is intended to eliminate, even though the database ends in the correct state.

Suggested fix — use `INSERT ... ON CONFLICT DO NOTHING`, or catch `IntegrityError`, roll back, re-read the defaults, and return success only after confirming that every expected resource now exists; re-raise if the post-rollback state is incomplete.

reana_db/models.py
tiborsimko: Addressed in commit `b773c39`, kept as a separate fixup on top of the original commit for review clarity. This fixup commit can be squashed into the original commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana_db/models.py
tiborsimko: Addressed in commit `b773c39`, kept as a separate fixup on top of the original commit for review clarity. This fixup commit can be squashed into the original commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana_db/cli.py
tiborsimko: Addressed in commit `b773c39`, kept as a separate fixup on top of the original commit for review clarity. This fixup commit can be squashed into the original commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana_db/models.py
```diff
@@ -163,12 +164,26 @@ class User(Base, Timestamp, QuotaBase):
...
+        ),
+        UniqueConstraint("idp_issuer", "idp_subject", name="uq_user__idp_identity"),
+        {"schema": "__reana"},
+    )
 
     id_ = Column(UUIDType, primary_key=True, unique=True, default=generate_uuid)
     email = Column(String(length=255), unique=True, primary_key=True)
     full_name = Column(String(length=255))
     username = Column(String(length=255))
+    idp_issuer = Column(String(length=255))
```
tiborsimko, line 179: **🤖 PR269-06: Periodic CPU quota updates exclude JWT-only users (MEDIUM)**

`update_users_cpu_quota(user=None)` still selects users by joining
`UserToken` and filtering for an active legacy token
(`reana_db/utils.py:653-657`). The new JIT OIDC users deliberately have no
`UserToken` row, so the periodic/all-users quota job silently omits every
JWT-only account. The disk updater in the same file already shows the intended
shape by using `Session.query(User).all()`.

The affected production chain is Helm's quota CronJob -> `reana-db quota
resource-usage-update` -> `update_users_cpu_quota(user=None)`. It loses both
CPU accounting and periodic-window advancement for OIDC users. The
workflow-termination path passes `user=workflow.owner` explicitly and
therefore works, which can mask the defect when testing a deployment that uses
termination-based rather than periodic accounting.

Suggested fix — select the intended account population independently of
`UserToken` (for example all users, or an explicit account-status/identity
predicate if disabled accounts must be skipped). Add coverage with a user that
has `idp_issuer`/`idp_subject` and no legacy token and assert both the bulk
updater and the periodic quota-window advancement process it.

reana_db/models.py
```diff
@@ -493,6 +508,10 @@ class InteractiveSession(Base, Timestamp, QuotaBase):
         nullable=False,
         default=InteractiveSessionType.jupyter,
     )
+    session_secret = Column(
```
tiborsimko: **🤖 PR269-07: Secret-key rotation omits both new encrypted columns (MEDIUM)**

PR269 adds encrypted `User.gitlab_webhook_secret` and
`InteractiveSession.session_secret` columns, but
`change_key_encrypted_columns()` in `reana_db/utils.py` still reads and
rewrites only `UserToken.token`. After an operator runs the documented
`migrate-secret-key` procedure, legacy tokens are encrypted with the new key
while existing webhook and notebook secrets remain ciphertext under the old
key. Once services restart with only the new key, those values cannot be
decrypted.

This can break active GitLab webhooks and access to running notebook sessions
precisely during a security-sensitive key rotation. There is no round-trip
rotation test for either new field. Once the legacy `UserToken` model is
removed, its `token` leg should also disappear from this utility rather than
remaining the only supported column.

Suggested fix — include every live `EncryptedType` column in one rotation
transaction, preserving `NULL` values, and add a test that creates all
currently supported secret types with the old key, rotates, discards ORM
state, switches to the new key, and successfully reloads/decrypts each value.
Cross-check this with PR789-23's removal of the remaining legacy-token
consumer.

reana_db/models.py
tiborsimko: **🤖 PR269-02 follow-up (2026-07-28): verified fixed**

*Component:* reana-db — auth persistence model tests

Focused model tests now cover duplicate `(idp_issuer, idp_subject)` rejection,
the complete/empty identity invariant, and encrypted store/commit/expunge/reload
round trips for `gitlab_webhook_secret` and `session_secret`.

Verified fixed at `b773c39`. PR269-07 separately covers rotation of those
encrypted values; it does not reopen their basic persistence coverage.

reana_db/models.py
tiborsimko: **🤖 PR269-03 follow-up (2026-07-28): verified fixed**

*Component:* reana-db — OIDC identity invariant

The model and migration now add the named `idp_identity_complete` check:
`(idp_issuer IS NULL) = (idp_subject IS NULL)`. Both fields are therefore
absent or present together, while the pair remains unique when populated.
Tests exercise both partial forms.

Verified fixed at `b773c39`. No residual.

reana_db/cli.py
tiborsimko: **🤖 PR269-04 follow-up (2026-07-28): verified fixed**

*Component:* reana-db — default resource initialisation

`Resource.initialise_default_resources` now accepts the working session,
handles a concurrent unique-name winner by rolling back and re-reading the
complete expected resource set, and re-raises unrelated/incomplete failures.
The race outcome is covered directly.

Verified fixed at `b773c39`. No residual.

reana_db/utils.py
```diff
@@ -650,12 +648,9 @@ def update_users_cpu_quota(user=None, override_policy_checks: bool = False) -> N
...
     else:
-        users = (
-            Session.query(User)
-            .join(UserToken)
-            .filter_by(status=UserTokenStatus.active)  # skip users with no active token
-            .all()
-        )
+        # JWT-authenticated users do not have a legacy UserToken row. Quotas
+        # are account properties, so maintenance must include every user.
+        users = Session.query(User).all()
```
tiborsimko: **🤖 PR269-08: Bulk CPU-quota maintenance becomes a full-row N+1 scan (LOW)**

Selecting every `User` fixes the correctness gap for JWT-only accounts, but the
periodic bulk path now materialises every column of every historical user and
then performs a separate `UserResource` lookup for each row. The work therefore
grows with all accounts ever provisioned, including accounts without a CPU
quota, and loads encrypted/profile columns that this task does not need before
doing the existing per-user workflow aggregation.

Suggested fix — drive the bulk path from CPU `UserResource` rows (joining or
batch-loading only the user fields actually needed) and keep the explicit
single-user path unchanged. This preserves legacy/unlinked users that really
hold quota while avoiding a full `user_` scan and one quota lookup per account.

reana_db/models.py
tiborsimko: **🤖 PR269-06 follow-up (2026-08-11): verified fixed at ac8bbfc**

Bulk CPU-quota maintenance now includes `User` rows independently of legacy
`UserToken` state, and the regression coverage includes a JWT-only account with
no token row. This closes the correctness gap for newly provisioned OIDC users.

reana_db/models.py
tiborsimko: **🤖 PR269-07 follow-up (2026-08-11): verified fixed at ac8bbfc**

Secret-key rotation now materialises and rewrites all encrypted token-bearing
columns: legacy `UserToken.token`, `User.gitlab_webhook_secret`, and
`InteractiveSession.session_secret`. The old key is confined to a `try/finally`
read phase and restored even on failure, and round-trip tests cover the added
columns. Rotating the database key no longer strands the new credentials.

reana_db/utils.py
Sohn123: Fixed in e6af5ef: bulk CPU-quota maintenance is now driven by `UserResource` rows (`filter_by(resource_id=cpu_resource.id_)`) instead of `Session.query(User).all()`. `UserResource.user_id` carries all the per-account identity the loop needs, so JWT-only/legacy accounts still get covered without a full `user_` table scan or a per-user lookup.

reana_db/alembic/versions/20260715_1200_972826f7f99f_require_complete_idp_identity.py
```diff
@@ -0,0 +1,70 @@
...
+    # database -- confirmed by running the real upgrade() against a live
+    # Postgres connection. This is *not* the same name a fresh install gets
+    # for the equivalent ``CheckConstraint`` in ``models.py``, which is
+    # declared on ``Base.metadata`` (naming_convention configured) and so
+    # is expanded to "ck_user__idp_identity_complete" despite also passing
+    # an explicit ``name=``. ``op.f("ck_user__idp_identity_complete")`` named
+    # a constraint that this migration's own upgrade() never created, so
+    # downgrading it always failed with "constraint does not exist".
+    op.drop_constraint(
+        "idp_identity_complete",
```
tiborsimko: **🤖 PR269-09: Alembic name guard ignores the configured naming convention (LOW)**

The new explanation and AST guard model `op.create_check_constraint()` as
creating the literal name `idp_identity_complete`. That is not how this
migration runs: `alembic/env.py` passes `Base.metadata` to
`context.configure()`, and its `ck_%(table_name)s_%(constraint_name)s` naming
convention is applied symmetrically to Alembic create and drop operations.

Rendering the real operations with that migration context produces
`ADD CONSTRAINT ck_user__idp_identity_complete` for the bare create. Both the
original `op.f("ck_user__idp_identity_complete")` drop and the amended bare
`"idp_identity_complete"` drop emit
`DROP CONSTRAINT ck_user__idp_identity_complete`. The original downgrade was
therefore already DDL-equivalent and the new comment's claimed failure mode is
incorrect.

The source-only test compares spellings before Alembic expands them, so it
would reject the correct `op.f(...)`/bare-name pair and can accept spellings
that render differently under a convention. It pins an AST property rather
than the database identifier that upgrade and downgrade actually use.

Suggested fix — remove the misleading comment and replace the AST-name guard
with a test that renders or executes upgrade/downgrade through REANA's real
Alembic environment and compares emitted/actual identifiers. Keep either the
original `op.f("ck_user__idp_identity_complete")` or the equivalent bare
semantic name, but test the resulting DDL rather than requiring identical
source strings.

reana_db/alembic/versions/20260715_1200_972826f7f99f_require_complete_idp_identity.py
```diff
@@ -0,0 +1,100 @@
...
+    ).scalar()
+    if new_name_exists:
+        # Already correctly named -- nothing to do. Not expected in normal
+        # Alembic operation (revision tracking prevents re-running an
+        # already-applied upgrade), but keeps this function idempotent under
+        # direct re-invocation too, matching the old-name branch above.
+        return
+
+    op.create_check_constraint(
+        "ck_user__idp_identity_complete",
```
tiborsimko: **🤖 PR269-10: Migration `972826f7f99f` creates a double-prefixed constraint name (MEDIUM)**

The amended migration passes the fully expanded name
`ck_user__idp_identity_complete` to `op.create_check_constraint()` on the
premise that this Alembic operation has no naming convention. REANA's Alembic
environment passes `Base.metadata` as `target_metadata` in both online and
offline mode, so Alembic propagates its
`ck_%(table_name)s_%(constraint_name)s` convention into the operation. Because
that template contains `%(constraint_name)s`, SQLAlchemy applies it to an
explicitly named check constraint as well.

Rendering the operation through that migration context gives:

```text
op.create_check_constraint("ck_user__idp_identity_complete", ...)
  -> ADD CONSTRAINT ck_user__ck_user__idp_identity_complete

op.create_check_constraint("idp_identity_complete", ...)
  -> ADD CONSTRAINT ck_user__idp_identity_complete
```

The created double-prefixed name does not match `create_all()`, the
`new_name_exists` idempotency guard, or either name dropped by `downgrade()`.
Downgrade therefore silently leaves the constraint behind, and a subsequent
upgrade attempts to create the same double-prefixed constraint again. The new
AST test approves the incorrect spelling because it compares the source
literal with the ORM's already-expanded name rather than comparing emitted DDL.

Suggested fix — use the semantic name `"idp_identity_complete"`, or mark the
expanded name as final with `op.f("ck_user__idp_identity_complete")`. Remove the
comments claiming the operation has no naming convention, and replace the
source-literal assertion with a test that renders or executes upgrade and
downgrade using `Base.metadata`, asserting the actual DDL identifiers.

reana_db/utils.py
```diff
@@ -860,26 +851,64 @@ def change_key_encrypted_columns(old_key):
...
+                Session.query(InteractiveSession.id_, InteractiveSession.session_secret)
+                .filter(InteractiveSession.session_secret.isnot(None))
+                .all()
+            )
+        except (InvalidPaddingError, ValueError, UnicodeDecodeError) as error:
+            raise RuntimeError(
+                "Could not decrypt one or more secrets with the old key. This "
+                "usually means a secret was created or renewed with the new "
+                "key after it was deployed but before this rotation ran. "
+                "Quiesce GitLab webhook and interactive-session writes (or "
```
tiborsimko: **🤖 PR269-11: Mixed-key rows make secret rotation permanently unretryable (MEDIUM)**

`change_key_encrypted_columns()` switches `DB_SECRET_KEY` to the old key and
materialises every encrypted value before writing anything. If a live server
has already written even one row with the new key, reading that row under the
old key raises and aborts the rotation. The new error message tells the
operator to quiesce writes and retry.

Quiescing prevents additional mixed-key rows, but it cannot change the
new-key row already stored in the database. Every retry still reads that row
with the old key and fails at the same point. No data is corrupted by the
failed transaction, but the deployment remains split between rows the newly
configured server can decrypt and old-key rows that the rotation command can
never reach. The regression test asserts the initial failure and unchanged
data, but does not demonstrate a successful recovery after following the
reported remediation.

Suggested fix — make the rotation understand mixed-key state, for example by
reading raw ciphertext and trying the old and new keys per value, re-encrypting
only values that decrypt with the old key. Alternatively, require and enforce
quiescence before deploying the new key and accept both keys explicitly during
one migration window. Add a test with simultaneous old-key and new-key rows
that follows the documented recovery procedure and completes successfully.

reana_db/version.py
```diff
@@ -14,4 +14,4 @@
 
 from __future__ import absolute_import, print_function
 
-__version__ = "0.95.0a11"
+__version__ = "0.95.0a10"
```
tiborsimko: **🤖 PR269-12: The branch reverts reana-db's released version and lowers two downstream pins (LOW)**

Commit `55cbf79` changes `reana_db/version.py` from `0.95.0a11` back to
`0.95.0a10`, undoing the version change made by master's
`chore(master): release 0.95.0a11 (#271)`. Against their respective merge
bases, the reana-server and reana-workflow-controller branches likewise lower
both of their `reana-db` requirements from `>=0.95.0a11` to `>=0.95.0a10`.

These edits appear to have been intended to keep the auth branches on the
latest published reana-db version when that version was still `0.95.0a10`.
Now that master has released `0.95.0a11`, however, they have become backwards
changes: merging PR269 would leave reana-db reporting a version lower than its
newest tag and release, while PR789 and PR696 would re-admit a dependency
version that their current bases deliberately excluded.

This is separate from deciding which future reana-db release will first carry
the new authentication models and migrations. The immediate concern is only
that this PR set should not regress versions and dependency floors already
present on master.

Suggested fix — restore `__version__ = "0.95.0a11"` and the four downstream
requirements to `reana-db>=0.95.0a11`, or drop these branch-specific hunks so
that each branch inherits its current base. The normal release procedure can
later raise the requirements to the first reana-db release containing the new
models. No corresponding reana-commons change is needed; that branch already
retains `0.95.0a22`.

reana_db/alembic/versions/20260715_1200_972826f7f99f_require_complete_idp_identity.py
tiborsimko: **🤖 PR269-09 follow-up (2026-08-20): still partial at 5527590**

The amended migration now spells the source literal as
`ck_user__idp_identity_complete`, but the underlying premise remains
incorrect. `alembic/env.py` passes `Base.metadata` to `context.configure()`, so
Alembic applies REANA's `ck_%(table_name)s_%(constraint_name)s` convention to
`op.create_check_constraint()`.

Rendering the current operation therefore produces:

```text
"ck_user__idp_identity_complete"
  -> ck_user__ck_user__idp_identity_complete
```

The correct bare semantic name, `"idp_identity_complete"`, renders as
`ck_user__idp_identity_complete`. The new AST guard still compares the source
literal with an already-expanded ORM name and consequently approves the wrong
operation while rejecting the correct semantic spelling. The requested
rendered-DDL coverage has not been added.

Suggested fix — replace the source-level name comparison with a test that runs
the operations through a migration context configured with
`target_metadata=Base.metadata` and asserts the emitted identifiers. Use the
bare semantic constraint name or an `op.f(...)` final name, and remove the
comments stating that the Alembic operation has no naming convention.

reana_db/alembic/versions/20260715_1200_972826f7f99f_require_complete_idp_identity.py
```diff
@@ -0,0 +1,103 @@
...
+        )
+
+    # A pre-naming-convention-fix run of this same revision may already have
+    # created the constraint under the literal name "idp_identity_complete"
+    # (before this migration was corrected to use the fully-expanded name).
+    # Alembic tracks revision IDs, not migration content, so any environment
+    # that already applied the old version of this revision will never re-run
+    # this function -- rename in place instead of erroring on a duplicate.
+    old_name_exists = connection.execute(
+        sa.text("SELECT 1 FROM pg_constraint WHERE conname = 'idp_identity_complete'")
```
tiborsimko: **🤖 PR269-13: An in-place migration repair cannot fix stamped databases and its probes are unscoped (LOW)**

Migration `972826f7f99f` now contains compatibility branches for earlier
constraint spellings. A database that already ran this revision is stamped
past it, however, so Alembic will never execute the amended `upgrade()` there.
This matters for developers who ran the intermediate branch revision that
created `ck_user__ck_user__idp_identity_complete`; the current repair cannot
remove or rename that stranded constraint.

The `old_name_exists` branch instead looks for the literal
`idp_identity_complete`, although the original bare semantic name was expanded
by REANA's naming convention to `ck_user__idp_identity_complete`. No released
form of this migration created the literal name. Both `pg_constraint` probes
also filter only on `conname`. A `new_name_exists` match on another table or
schema causes an early return and can silently leave `__reana.user_` without
the intended check constraint. A false `old_name_exists` match instead runs
`ALTER TABLE __reana.user_ RENAME CONSTRAINT` for a constraint that is not on
that table, aborting the migration.

Suggested fix — put compatibility repair in a new revision after the current
head so already-stamped databases execute it. Scope catalogue checks to the
`__reana.user_` relation, handle the intermediate double-prefixed name
explicitly, and test upgrades from each known stamped schema state. The later
GCM conversion correctly uses a new revision for this same reason and should
remain separate.

reana_db/models.py
```diff
@@ -493,6 +544,10 @@ class InteractiveSession(Base, Timestamp, QuotaBase):
...
+        EncryptedType(String(length=255), _secret_key, AesEngine, "pkcs5"),
```
tiborsimko: **🤖 PR269-14: Interactive-session bearer secrets remain on deterministic CBC encryption (LOW)**

The amendment moves `User.gitlab_webhook_secret` from deterministic
`AesEngine` encryption to authenticated, randomised `AesGcmEngine` encryption
because equality-leaking ciphertext is unnecessary for that capability.
`InteractiveSession.session_secret`, added by the same PR, remains on
`AesEngine` with deterministic AES-CBC.

A session secret is likewise a live bearer credential, is never queried by
ciphertext, and has no ciphertext uniqueness requirement. There is therefore
no functional reason to retain determinism, while equal plaintexts remain
visible as equal ciphertexts and CBC provides no authenticated-encryption
integrity guarantee.

Suggested fix — add a new migration that decrypts existing session secrets
with CBC and re-encrypts them with GCM, then change the model column to
`AesGcmEngine`. Keep the conversion in a new revision so already-stamped
databases receive it, and cover round-trip and key-rotation behaviour. As a
small related cleanup, the `"pkcs5"` argument can be removed from GCM column
configuration because `StringEncryptedType` does not apply CBC padding to the
sibling `AesGcmEngine` class.

reana_db/utils.py
tiborsimko: **🤖 PR269-08 follow-up (2026-08-21): verified fixed at 7c79195**

Bulk CPU-quota maintenance is now driven by CPU `UserResource` rows rather
than materialising every historical `User`. The explicit single-user path is
unchanged, and the bulk query selects only accounts that actually carry the
resource being maintained. This removes the reported full-row N+1 scan while
retaining linked and unlinked quota holders. No residual concern remains for
this finding.

reana_db/alembic/versions/20260715_1200_972826f7f99f_require_complete_idp_identity.py
tiborsimko: **🤖 PR269-09 follow-up (2026-08-21): verified fixed at 7c79195**

The source-level AST assertion has been replaced by a test that renders the
operations through a real Alembic `MigrationContext` using REANA's metadata
naming convention. It now checks the emitted database identifiers rather than
assuming that source literals are final names. The misleading explanation was
removed. No residual concern remains for this finding.

reana_db/alembic/versions/20260715_1200_972826f7f99f_require_complete_idp_identity.py
tiborsimko: **🤖 PR269-10 follow-up (2026-08-21): verified fixed at 7c79195**

The migration again passes the bare semantic name
`idp_identity_complete`, which REANA's naming convention expands exactly once
to `ck_user__idp_identity_complete`. The real-context DDL test covers upgrade
naming and rejects the former double-prefixed form. The downgrade uses raw SQL
that this test does not inspect, so this reply does not claim downgrade
coverage. Compatibility for databases already stamped by an intermediate form
of the revision is likewise separate from the fresh-execution naming defect.
No residual concern remains within this finding's scope.

reana_db/utils.py
tiborsimko: **🤖 PR269-11 follow-up (2026-08-21): verified fixed at 7c79195**

Secret rotation now reads each locked row by trying the old key first and the
new key second. Values already written or rotated with the new key are kept,
old-key values are re-encrypted, and only ciphertext valid under neither key
fails with an actionable error. Row locks prevent a concurrent new-key write
from being overwritten, and the error path rolls back pending work. The tests
exercise mixed-key and concurrent-writer recovery through real PostgreSQL
sessions. No residual concern remains for this finding.

reana_db/version.py
tiborsimko: **🤖 PR269-12 follow-up (2026-08-21): verified fixed at 7c79195**

The branch again reports reana-db `0.95.0a11`, and the reana-server and
reana-workflow-controller dependency floors have been restored to
`reana-db>=0.95.0a11`. The PR set no longer reverses versions already present
on master. Later release-specific bumps remain part of the normal release
procedure and are outside this finding. No residual concern remains.

reana_db/alembic/versions/20260715_1200_972826f7f99f_require_complete_idp_identity.py
Sohn123: Fixed as suggested: new forward migration (`20260824_1015_..._repair_idp_identity_constraint.py`, chained after the GCM-hardening migration) rather than editing the already-released `972826f7f99f`, so already-stamped databases actually execute the repair. Its constraint lookup is now scoped to `__reana.user_` specifically (joins `pg_class`/`pg_namespace`, filters on schema and table name), closing the unscoped-probe half too.

reana_db/models.py
Sohn123: Fixed, and generalized rather than patched in place. Added a shared `reana_db/secrets.py` primitive (`bearer_secret_column()`, always AES-GCM) and moved both `User.gitlab_webhook_secret` and `InteractiveSession.session_secret` onto it, with a new migration converting existing CBC-encrypted session secrets. Also added a regression test asserting no column in `models.py` declares `EncryptedType` directly outside this factory, so a future secret column can't silently reintroduce a deterministic engine the way this one did.

tiborsimko (commented):
**🤖 PR269-15: `Workflow.get_owner_access_token()` calls the removed `User.access_token` (LOW)**

The opaque-token model and `User.access_token` property are removed in this PR,
but the workflow helper which exposes that credential remains:

```python
def get_owner_access_token(self):
    """Return workflow owner access token."""
    ...
    return owner.access_token
```

Any future caller now receives `AttributeError`. There are no current runtime
callers, which is why the regression is not exercised by the test suite, but
leaving a public model helper with a deleted implementation is a trap for code
looking for the workflow owner's authentication credential.

Suggested fix — remove `Workflow.get_owner_access_token()` together with the
rest of the opaque-token surface so the model no longer advertises a credential
which cannot exist.

tiborsimko (commented):
**🤖 PR269-15 follow-up (2026-08-25): verified fixed at 74f2d31**

The obsolete `Workflow.get_owner_access_token()` method has been removed with
the deleted opaque-token model surface. There is no longer a public-looking
helper which calls the nonexistent `User.access_token` attribute. No residual
concern remains for this finding.

reana_db/alembic/versions/20260824_1015_3f6c0d2a1b7e_repair_idp_identity_constraint.py
```diff
@@ -0,0 +1,154 @@
...
+        "quota_period_months_positive",
+        "user_resource",
+        "quota_period_months IS NULL OR quota_period_months > 0",
+        schema="__reana",
+    )
+
+
+def upgrade():
+    """Converge every known stamped state on the ORM's constraint name."""
+    _repair_quota_constraint()
```
tiborsimko, line 120: **🤖 PR269-17: The quota repair was inserted into an already-stampable migration (LOW)**

Revision `3f6c0d2a1b7e` was added before this amendment and can already be
recorded in development databases. The new commit inserts the quota repair
into that existing revision:

```python
def upgrade():
    """Converge every known stamped state on the ORM's constraint name."""
    _repair_quota_constraint()
    ...
```

Any database stamped at `3f6c0d2a1b7e` or a later revision before this edit
will never execute `_repair_quota_constraint()`. It retains the drifted quota
constraint indefinitely even though a fresh database running the same branch
head is repaired. This is precisely the migration-history hazard described by
this revision's own module docstring:

> Editing that old revision cannot repair a database whose Alembic stamp is
> already past it, so compatibility belongs in this new, forward revision.

The revision has been present on the PR branch since commit `0453f9c`, so this
is reachable for local and end-to-end test databases even though the migration
has not been released.

Suggested fix — move the quota repair into a new forward migration after the
current branch head, keeping it idempotent across the corrected,
double-prefixed, literal, and missing states. Add a test which starts with an
Alembic stamp beyond `3f6c0d2a1b7e` and verifies that upgrading through the new
revision performs the repair.

tiborsimko (commented):
**🤖 PR269-16: The quota downgrade still targets the pre-repair constraint name (LOW)**

This PR does not change the quota migration — `eb240ac` correctly restores it
to the form already released in 0.95.0a10/a11 — but the new repair revision
makes its existing `downgrade()` unusable.

```python
op.drop_constraint(
    "ck_user_resource_quota_period_months_positive",
    "user_resource",
    type_="check",
    schema="__reana",
)
```

This repository's naming convention is
`ck_%(table_name)s_%(constraint_name)s`. Unless a name is marked as already
final, Alembic expands the string above a second time and emits a drop for
`ck_user_resource_ck_user_resource_quota_period_months_positive` — which is
exactly the name the released migration created, so this downgrade matched
until now.

The new forward repair does the opposite: it renames the double-prefixed or
literal form to `ck_user_resource_quota_period_months_positive`. A database
which has run that repair therefore fails when downgrading past
`06dbbeef6d9b`, because the downgrade asks PostgreSQL to remove a constraint
which no longer exists. Databases which skipped the in-place repair can have a
different historical name, so hard-coding only the corrected form is not
enough for every supported branch state either.

Suggested fix — make the downgrade inspect `pg_constraint` for the corrected,
double-prefixed, and literal compatibility names and drop the one that exists,
as the identity-constraint downgrade already does. Alternatively use
`op.f(...)` for the converged final name while retaining compatibility handling
for older stamped states. Add upgrade-then-downgrade tests for each known name.
Fixing this belongs with the repair work in this PR, since the repair is what
introduces the mismatch.

tiborsimko (approved):

